Issue #11, opened 2019-01-18
Fix other Platforms


m4xw:
**Functional**:
- Libnx Target
- Android aarch64/armv7 Target
- Windows x64/x86
- Raspberry aarch32/armv7
- Linux x86/x64, arm7neonhf

**Notes**:
Some changes for awk are needed across all platforms and Android doesn't generate the header properly yet (works if pre generated)
These could be generated and committed, but this might cause Issues with different toolchain versions being used.

"mupen_next" is our master branch for the time being

[More infos added later on...]

IntelMiner:
Pulled from bug report in other repository

I managed to build Retroarch + Mupen64 from git on an RPi 3, with the "[use new vendor library](https://github.com/libretro/mupen64plus-libretro/pull/39)" patches @m4xw requested I test

However running it seems to crash upon loading a game

The last lines of the output log are 

```
[INFO] [GL]: Found GL context: kms
[INFO] [GL]: Detecting screen resolution 1366x768.
[INFO] [EGL] Found EGL_EXT_platform_base, trying eglGetPlatformDisplayEXT
[INFO] [EGL]: EGL version: 1.4
[ERROR] [EGL]: #0x3009, EGL_BAD_MATCH
[ERROR] [Video]: Cannot open video driver ... Exiting ...
[ERROR] Fatal error received in: "init_video()"

```

Annoyingly looking up `#0x3009, EGL_BAD_MATCH` or `eglGetPlatformDisplayEXT` seem to bring up either source code, or Android development posts

The full RetroArch debug output is attached below as well

[crash.log](https://github.com/libretro/mupen64plus-libretro-nx/files/2791165/crash.log)

IntelMiner:
Looking things over. I may have grabbed the wrong end of the wrong stick, compiling Mupen64 from the upstream libretro repository, instead of the mupen64-libretro-nx fork

I'll rebuild the nx core fork over the weekend and report back

m4xw:
Yea back then I wanted you to test normal mupen.
We focus on m64p-next now, try first without the patch (its not in next yet) and then with.
If the patch breaks it for you, you probably don't have your system up2date.

IntelMiner:
I've got the latest firmware and the 4.14 kernel all ready to go, so that's all done and dusted

[...]

PR #14, opened 2019-01-27
Makefile: fix build on linux 64-bit

gvbr:
Able to compile after using `-f elf64` on my 64-bit linux machine
#11


m4xw:
Thanks!

m4xw:
@gvbr Dynarec works fine on x64 Linux?
Can you confirm that for me?

PR #16, opened 2019-01-29
Makefile: fix build for arm7neonhf platform

hhromic:
This platform is used by the libretro-super buildbot.

Tested using the following make invocation:

    make platform=unix-armv7-hardfloat-neon -j4 WITH_DYNAREC=arm

With this PR, the following line can be now added to the recipes in `recipes/linux/cores-linux-arm7neonhf`  of the libretro-super repository:

    mupen64plus_next libretro-mupen64plus_next https://github.com/libretro/mupen64plus-libretro-nx.git mupen_next YES GENERIC_GL Makefile . WITH_DYNAREC=arm

Note: I only tested that the build succeeds, but haven't tested the actual core in retroarch. However the build is very similar to the old mupen64plus core, should be fine. Nevertheless this PR is a good starting point to start receiving more feedback using the buildbot binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,26 @@ ifneq (,$(findstring unix,$(platform)))
       ASFLAGS = -f elf -d ELF_TYPE
    endif
 
+   ifneq (,$(findstring armv,$(platform)))
+      CPUFLAGS += -DARM -marm
+      ifneq (,$(findstring cortexa8,$(platform)))
+         CPUFLAGS += -mcpu=cortex-a8
+      else ifneq (,$(findstring cortexa9,$(platform)))
+         CPUFLAGS += -mcpu=cortex-a9
+      else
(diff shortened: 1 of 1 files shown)


m4xw:
Hey, thanks!
I assume you tested it with mupen_next? Because the quoted buildbot line points to master, which is old mupen.

hhromic:
Yes, sorry I made a mistake in the PR description (I typed that manually, not copy/pasted). I can confirm that I tested using the `mupen_next` branch, not the `master`.

This is the libretro-super invocation used after adding the recipe (temporarily pointing to my PR):

    SINGLE_CORE=mupen64plus_next CORE=mupen64plus_next ./libretro-buildbot-recipe.sh recipes/linux/cores-linux-arm7neonhf

PR #18, opened 2019-02-03
Makefile: add missing CFLAGS for C standard selection

hhromic:
Some feedback from RetroPie users already arrived!
A user discovered that there is a missing flag for older compilers, i.e. in Raspbian Jessie.

The `libnx` part is not strictly necessary because on newer compilers it defaults to `gnu11` already, nevertheless I explicitly added it for consistency.

---

In older versions of GCC, the `gnu11` standard is not selected by default.

Setting the standard to `gnu11` fixes these compilation errors:

    mupen64plus-core/src/device/r4300/new_dynarec/arm/assem_arm.c:4300:9: error:
    ‘for’ loop initial declarations are only allowed in C99 or C11 mode
         for(int hr=0;hr<HOST_REGS;hr++) {
         ^
    mupen64plus-core/src/device/r4300/new_dynarec/arm/assem_arm.c:4300:9: note:
    use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ else ifeq ($(platform), libnx)
    CPUOPTS := -g -march=armv8-a -mtune=cortex-a57 -mtp=soft -mcpu=cortex-a57+crc+fp+simd
    PLATCFLAGS = -O3 -ffast-math -funsafe-math-optimizations -fPIE -I$(PORTLIBS)/include/ -I$(LIBNX)/include/ -ffunction-sections -fdata-sections -ftls-model=local-exec -specs=$(LIBNX)/switch.specs
    PLATCFLAGS += $(INCLUDE) -D__SWITCH__=1 -DSWITCH -DHAVE_LIBNX -D_GLIBCXX_USE_C99_MATH_TR1 -D_LDBL_EQ_DBL -funroll-loops
-   CXXFLAGS += -fno-rtti -std=gnu++11
+   CXXFLAGS += -fno-rtti
    COREFLAGS += -DOS_LINUX
    GLES = 0
    WITH_DYNAREC = aarch64
@@ -341,9 +341,9 @@ endif
    CXXFLAGS += -fvisibility-inlines-hidden
 endif
 
-ifneq ($(platform), libnx)
-CXXFLAGS += -std=c++11
-endif
+# set C/C++ standard to use
+CFLAGS += -std=gnu11
+CXXFLAGS += -std=gnu++11
 
 ifeq ($(PIC), 1)
    fpic = -fPIC


hhromic:
Oops I just had a chance to try it on my Raspberry Pi and is failing to build.
I should have tested it before submitting the PR, apologies, please don't merge it yet.

hhromic:
Alright, now everything is building fine. I also aligned the `CXXFLAGS` to use `gnu++11` instead of `c++11` and the build is much cleaner now :)

The only warning remaining is this one:
```
mupen64plus-core/subprojects/minizip/zip.c: In function ‘zipOpenNewFileInZip4_64’:
mupen64plus-core/subprojects/minizip/zip.c:1249:28: warning: assignment from incompatible pointer type [-Wincompatible-pointer-types]
         zi->ci.pcrc_32_tab = get_crc_table();
                            ^
```
Before with `c++11` I was getting some more warnings, therefore definitively the current codebase is meant to be built using the `gnu11` and `gnu++11` standards for C/C++ respectively.

Btw, I have some game-related feedback from users in the RetroPie forum, would you like me to collate that into a separate GitHub issue?  The feedback seems very positive, a user even claims some games run a bit better than in the standalone mupen64plus provided by RetroPie.

m4xw:
@hhromic Hm, I think a "feedback issue" should happen after GlideN64 update, not before.
But sure, open a informational issue "Mupen64Plus-Next Initial Feedback"

hhromic:
Sure, I also agree it would be more useful after the GlideN64 update as this is where the core is heading to.
Another user was concerned that the new updates could actually bring problems to the Rasberry Pi due to its GPU not being very powerful, and in that case the older versions simply work better in some cases. So definitively I would like to hear his experience after the new GlideN64.

m4xw:
@hhromic What is the state of this PR?

hhromic:
@m4xw tested and ready to merge, sir !

m4xw:
@hhromic Can I ask you to remove the gnu stuff from the libnx target and have it set unconditionally?
Seems to be the right thing to do?

hhromic:
You refer to remove setting the std in this block:
```make
   PLATCFLAGS += $(INCLUDE) -D__SWITCH__=1 -DSWITCH -DHAVE_LIBNX -D_GLIBCXX_USE_C99_MATH_TR1 -D_LDBL_EQ_DBL -funroll-loops
   CFLAGS += -std=gnu11
   CXXFLAGS += -fno-rtti -std=gnu++11
   COREFLAGS += -DOS_LINUX
```

And instead consolidate it here (ofc removing the conditional) so they are always set?
```make
ifneq ($(platform), libnx)
CFLAGS += -std=gnu11
CXXFLAGS += -std=gnu++11
endif
```

If so, yes that sounds much better to me and I agree is the better/robust thing to do.

m4xw:
Yea just leave the "-fno-rtti" and set it for everything

hhromic:
Okay, I will double check if `gnu11`/`gnu++11` are available in MSYS2 so the buildbot doesn't screw up and then update the PR accordingly.

m4xw:
yolo ¯\\\_(ツ)_/¯

hhromic:
Okay, tested and building fine in RPI and Windows, so I guess the others should be fine too.

m4xw:
We will find out ;)
While MSVC is technically supported too, it's not in use for this core, so thats fine with me.

PR #21, opened 2019-02-07
libretro interface improvements

### LINKED ISSUE #15: Core options prefix
I noticed that this core is using the same core options prefix than the older/regular `mupen64plus-libretro` core. I know this core is supposed to eventually replace the older one, but it is possible that in the meantime you differentiate the options? for example by appending `nx` to the prefix.

This is important to ensure this core plays nicely alongside the older core when both are installed at the same time, a very likely situation given the status of this core as for testing. This is specially important to make sure users don't get confused and are not facing errors because of this core taking options set for the older core.

hhromic:
* define `CORE_NAME` and use to prefix core options (closes #15)
* minor code style cleanups
* simplify code using conditional operators

Tested build in Linux64 and Windows64.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -516,20 +516,14 @@ void update_variables()
     var.value = NULL;
     if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
     {
-        if (!strcmp(var.value, "HLE"))
-            rspMode = 0;
-        else
-            rspMode = 1;
+        rspMode = !strcmp(var.value, "HLE") ? 0 : 1;
     }
 
     var.key = CORE_NAME "-BilinearMode";
     var.value = NULL;
     if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
     {
-        if (!strcmp(var.value, "3point"))
-            bilinearMode = 0;
-        else
-            bilinearMode = 1;
+        bilinearMode = !strcmp(var.value, "3point") ? 0 : 1;
     }
 
     var.key = CORE_NAME "-MultiSampling";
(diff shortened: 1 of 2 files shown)


hhromic:
Example usage of the `CORE_NAME` define:
```
$ make
(...)
$ strings mupen64plus_next_libretro.dll | grep \\-Enable
mupen64plus-EnableFBEmulation
mupen64plus-EnableLODEmulation
mupen64plus-EnableCopyColorToRDRAM
mupen64plus-EnableCopyDepthToRDRAM
mupen64plus-EnableHWLighting
mupen64plus-EnableNativeResTexrects
mupen64plus-EnableLegacyBlending
mupen64plus-EnableFragmentDepthWrite
mupen64plus-EnableShadersStorage

$ CFLAGS="$CFLAGS -DCORE_NAME=\\\"mupen64plus-next\\\"" make
(...)
$ strings mupen64plus_next_libretro.dll | grep \\-Enable
mupen64plus-next-EnableFBEmulation
mupen64plus-next-EnableLODEmulation
mupen64plus-next-EnableCopyColorToRDRAM
mupen64plus-next-EnableCopyDepthToRDRAM
mupen64plus-next-EnableHWLighting
mupen64plus-next-EnableNativeResTexrects
mupen64plus-next-EnableLegacyBlending
mupen64plus-next-EnableFragmentDepthWrite
mupen64plus-next-EnableShadersStorage
```

libretro/libretro.c
```diff
@@ -312,7 +316,7 @@ static void emu_step_initialize(void)
 
 static void EmuThreadFunction(void)
 {
-    log_cb(RETRO_LOG_INFO, "EmuThread: M64CMD_EXECUTE. \n");
+    log_cb(RETRO_LOG_INFO, "EmuThread: M64CMD_EXECUTE\n");
```
m4xw: Can you prefix this, while you are at it?

libretro/libretro.c
```diff
@@ -508,94 +512,70 @@ void update_variables()
...
-    var.key = "mupen64plus-rspmode";
+    var.key = CORE_NAME "-rspmode";
     var.value = NULL;
     if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
     {
-        if (!strcmp(var.value, "HLE"))
-            rspMode = 0;
-        else
-            rspMode = 1;
+        rspMode = strcmp(var.value, "HLE") ? 1 : 0;
```
m4xw: Honestly not sure if I like these, for a experienced programmer its no Issue, but it can be confusing for others.
It's fine with me tho.

libretro/libretro.c
```diff
@@ -803,48 +756,38 @@ void update_variables()
         }
     }
 
-    if (AspectRatio == 1)
-        var.key = "mupen64plus-43screensize";
-    else
-        var.key = "mupen64plus-169screensize";
+    var.key = AspectRatio == 1 ? CORE_NAME "-43screensize" : CORE_NAME "-169screensize";
```
m4xw: I prefer it like this for readability
`var.key = (AspectRatio == 1 ? CORE_NAME "-43screensize" : CORE_NAME "-169screensize");`

libretro/libretro.c
```diff
@@ -918,7 +859,7 @@ static void format_saved_memory(void)
 static void context_reset(void)
 {
     static bool first_init = true;
-    printf("context_reset.\n");
+    log_cb(RETRO_LOG_INFO, CORE_NAME ": context reset\n");
```
m4xw: Not like it matters much, but call it context_reset

m4xw:
Thanks btw ;)

libretro/libretro.c
hhromic: sure, will do.

libretro/libretro.c
hhromic: I think we should level programmers up and not down :) The conditional operator is a very elegant tool IMHO and programmers should be more aware of them.

I agree that in this particular case `strcmp()` can be a bit confusing because it has an "inverse" logical meaning, when the strings are equal you get `0` ("false") instead of `!=0` ("true"). Perhaps I can add back the negation operator (`!`) so it reads more naturally:

    rspMode = !strcmp(var.value, "HLE") ? 0 : 1;

what do you think?

libretro/libretro.c
hhromic: sure no problem.

libretro/libretro.c
m4xw: I like this more, and yes I have this particular thought because of what you cited.

libretro/libretro.c
hhromic: How about doing the right thing and using `RETRO_LOG_DEBUG` for these very low-level mesages and then use the function names?

libretro/libretro.c
hhromic: sure then, I will add the negations so they read more naturally :)

libretro/libretro.c
m4xw: Sure

libretro/libretro.c
hhromic: Now that I was doing this little refactoring, I have a question.
In this code block (as is currently):
```c
var.key = CORE_NAME "-txFilterIgnoreBG";
if (!strcmp(var.value, "True"))
    txFilterIgnoreBG = 0;
else
    txFilterIgnoreBG = 1;
```
Shouldn't it be the other way around? True=Ignore, False=Don't ignore? is this correct?

hhromic: I answer myself, the option is defined as this:
```
      { "mupen64plus-txFilterIgnoreBG",
            "Filter background textures; True|False" },
```
Meaning that indeed True=Filter (don't ignore), False=DontFilter (do ignore)
Was a bit confusing :D

hhromic:
I applied your comments now, let me know if you are happy with them.
And thanks for being open to contributions, appreciated!

m4xw:
Looks good on a first look, I will check it more indepth in a bit, busy with work rn

hhromic:
Sure, take your time, no rush.

hhromic:
@m4xw added support for easily setting CORE_NAME for compilation (optional)

Allows for more elegantly doing:

    make CORE_NAME=customname"

Instead of the cumbersome:

    CFLAGS="$CFLAGS -DCORE_NAME=\\\"customname\\\"" CXXFLAGS="$CXXFLAGS -DCORE_NAME=\\\"customname\\\""make

Tested and working fine.

m4xw:
@hhromic Is this ready?

hhromic:
Yes ready and tested, should be fine. The latest commit was just a convenience that I took the opportunity to slip in this PR :)

m4xw:
Now grab yourself a beer! Cheers

PR #23, opened 2019-02-11
Makefile(android): set the C standard for building

hhromic:
Let's leave this fixed so we don't forget later.
Refs:
https://github.com/libretro/mupen64plus-libretro-nx/issues/11#issuecomment-462199279
https://github.com/libretro/mupen64plus-libretro-nx/issues/11#issuecomment-462340511


m4xw:
Did you try if that works?
I've set it before and it would still use c89

hhromic:
No, I was trusting the reporter's word (also makes sense for it to work):
> Yeah, changing the line works.

I can set an Android build env later during the week and test it myself if you prefer, np.

m4xw:
I assume he meant replacing the `for`

hhromic:
Umm, you are right, it is ambiguous now for me. Well I'll give it a try myself anyway, just keep the PR open then :)

hhromic:
@m4xw finally I got some time to test this myself.

I setup a build environment that is the same as the libretro buildbot. I understand now what is going on with the `LOCAL_CPPFLAGS` and `LOCAL_CFLAGS` shenanigans for Android makefiles.

For the record, it turns out that `LOCAL_CFLAGS` is actually applied to both C *and* C++ source code compilation. There is another variable `LOCAL_CONLYFLAGS` specifically for C-only source code.

Therefore, adding `-std=gnu11` to `LOCAL_CFLAGS` like common-sense would suggest, allows it to build but the C++ compiler starts to complain a lot that this standard is not supported (logically). However when adding the flag to the correct `LOCAL_CONLYFLAGS` variable, everything builds without a fuzz.

* Tested for `APP_ABI=arm64-v8a` and still is building fine.
* Tested for `APP_ABI=armeabi-v7a` and now builds without errors (with a temporary `asm_defines_gas.h`). Of course, still `asm_defines_gas.h` needs to be generated, but that's for another PR.

This PR is now tested and ready to merge.

Ref: https://android.googlesource.com/platform/build/+/7429e21%5E%21/
Ref:

m4xw:
@hhromic Can you generate the asm defines for now for the arm toolchain and commit it in the assem folder?
Right now if the platform generates them they will pop up in asm_defines/ and that takes priority over the backend folders, so lets just commit them for the time being.
Also armv7 should be fine to add to supported ABI's after that!

hhromic:
@m4xw sure thing. However I have never done such thing before. My google-fu is quite decent but I would appreciate some pointers (files to check, scripts). I can take it from there. A good pointer for me would be where to look on how is done for other targets. I understand there is an awk script for this?

hhromic:
> Also armv7 should be fine to add to supported ABI's after that!

This should be done here or in the libretro-super repo?

m4xw:
basically just compile until it throws a error, then:
`strings "mupen64plus-core/src/asm_defines/asm_defines.o" | tr -d '\r' | awk -v dest_dir="./mupen64plus-core/src/asm_defines" -f ./mupen64plus-core/tools/gen_asm_defines.awk`

m4xw:
> This should be done here or in the libretro-super repo?

Needs to be added to Application.mk APP_ABI first

hhromic:
> basically just compile until it throws a error, then:
`strings "mupen64plus-core/src/asm_defines/asm_defines.o" | tr -d '\r' | awk -v dest_dir="./mupen64plus-core/src/asm_defines" -f ./mupen64plus-core/tools/gen_asm_defines.awk`

That is exactly what I was thinking on doing, just needed that shell snippet!  Will do.

> Needs to be added to Application.mk APP_ABI first

Just now figured out the same :) Will do too after testing all works.

hhromic:
Alrighty, added the ARM assembler defines now, and enabled the armeabi-v7a ABI too.
Tested the builds for Linux64, Win64 and Android(armv7/armv8). So far nothing seems broken and Android armv7 builds fine now.

hhromic:
The build for android armv7 seems to be working fine now :)
https://buildbot.libretro.com/nightly/android/latest/armeabi-v7a/

hhromic:
But in discord it was reported the core crashes :(, so I guess we have to keep this as WIP.
I will try to setup a testing env on my android 32 bits device for later.

m4xw:
I smell layer 8

hhromic:
Hopefully :)  When I get a chance to test it myself on a real device will confirm it working or not :)

m4xw:
confirmed it, hmm.

hhromic:
Yup, I confirm it now too. It's crashing, old mupen64plus is not crashing.
I will try to gather a log to see if it's something trivial. If you are interested on making this work on android armv7, I can open a new issue to track progress in there instead.

PR #24, opened 2019-02-14
libretro-common: update retro_miscellaneous include from upstream

hhromic:
* Silences warning about `PRI_SIZET` being incorrectly set for the `size_t` type in non-windows 32 bits platforms.

Minor harmless update, but anyway tested too.


m4xw:
Can't do this, will do a subrepo pull later.

hhromic:
Sure, it's already in upstream so whenever you subrepo pull, will be integrated.
It just silences warnings, so no rush.

m4xw:
Yea that's why (would cause conflicts), if it was a new change it would be different (depending on the change ofc).
But thanks anyway!

hhromic:
Yes, apologies I just realised that `libretro-common` is actually configured as git subrepo folder, didn't notice it before :)  So agree should def be updated via subrepo.

PR #32, opened 2019-03-01
Add configurable max texture cache size

### LINKED ISSUE #28: Instability/Corruption likely related to Texture cache size on low end devices.
Hello,
I've noticed that on RetroPie after 20 minutes or so the emulator starts glitching and then it crashes.
I tried disabling the framebuffer emulation, but the problem remains.

hhromic:
Set default to `1500` for VideCore, `4000` for LibNX and `8000` for others.

Values backported from upstream:
https://github.com/gonetz/GLideN64/blob/master/src/Textures.h#L103-L107

Users can select any max cache size they require for their devices without re-compiling.
Fixes #28

### DIFF
--- a/GLideN64/src/Textures.cpp
+++ b/GLideN64/src/Textures.cpp
@@ -18,6 +18,8 @@
 #include "GLideNHQ/Ext_TxFilter.h"
 #include "TextureFilterHandler.h"
 
+#include "../../../custom/GLideN64/GLideN64_libretro.h"
+
 #ifdef HAVE_LIBNX
 #include <switch.h>
 #endif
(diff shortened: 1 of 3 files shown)


hhromic:
Let me know if you prefer to keep the old values (1500, 4000, 16384) instead of the upstream ones.

hhromic:
Renamed the option from `GlobalTxCacheSize` to `MaxTxCacheSize` (better IMHO).

m4xw:
I was about to press the big green button :P 
Is it ready?

hhromic:
Yes yes, sorry for that :)  I realised too late that "global" wasn't a good name for it and changed it.

m4xw:
Thanks

PR #45, opened 2019-04-06
Add LLE settings.

mudlord:
Start importing of settings.

Haven't worked out yet how to distinguish between the two render plugins.

Why not have a selection to pick each plugin and then it picks the right RSPs accordingly?

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -88,6 +88,7 @@ static unsigned initial_boot        = true;
 static unsigned audio_buffer_size   = 2048;
 
 static unsigned retro_filtering     = 0;
+static unsigned retro_dithering = 0;
 static bool     first_context_reset = false;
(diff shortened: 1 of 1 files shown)


m4xw:
> Why not have a selection to pick each plugin and then it picks the right RSPs accordingly?

That was the plan, show core options depending on RSP/RDP selection, anyway 5am soon so will check that over the weekend.
You might need to remind me ;)

mudlord:
Thats a good idea, but I am wondering does the libretro API allow for such things?

Or will it need to be on RetroArch's side?

m4xw:
It needs to have the same amount of options, then we could do it (swap them on the fly)

m4xw:
merged for now, the branch is not used anywhere yet, so we can figure something later.

PR #53, opened 2019-05-21
Adding the 4:3 resolution of 1440x1080

Gagert:
On 1080p monitors this 4:3 resolution of 1440x1080 is quite beneficial compared to it's closest neighbor 1280x960, because this resolution gets rid of the black borders on the upper- and lower part of the image itself.


m4xw:
👌 ty

PR #64, opened 2019-05-27
Added PS Classic

evilwraith:


GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
```diff
@@ -35,9 +35,15 @@ typedef double GLdouble;
 #undef GL_COPY_READ_BUFFER_BINDING
 #undef GL_COPY_WRITE_BUFFER_BINDING
 #include <GL/glcorearb.h>
+
+#elif defined(CLASSIC)
```
m4xw: Using GLES2/3 will define EGL, thus this will never be triggered (or at least shouldn't be)

m4xw: https://github.com/libretro/mupen64plus-libretro-nx/blob/GLideN64/Makefile.common#L366-L384

PR #74, opened 2019-07-15
Makefile: added devices for Amlogic & Rockchip platforms

5schatten:
The purpose of this PR is to add platforms for Amlogic & Rockchip while also build the core for specific devices with different ARM cores (A53/A17/A72/A73) & OpenGL ES 2.0 or 3.0 support. The Amlogic platform names where derived from the SoC naming scheme used by [meson-linux](http://linux-meson.com/doku.php#supported_wip_soc_families) and LibreELEC.

- Amlogic S905/S905X/S912 (GX) / S905X2 (G12A) & S922X/A311D (G12B)
- Rockchip RK3288/RK3328/RK3399
- updated `GLFunctions.cpp` to define `-DUSE_GENERIC_GLESV2` instead `-DAMLOGIC` because on most platforms with `libGLESv2.so` support there should be a lib in `/usr/lib/...` or a symlinkt to the vendor blob like `libmali.so` which can be used for Mali GPUs.
- removed `UNDEF_GL_GLEXT_PROTOTYPES` from `GLFunctions.h`

@shantigilbert have a look at this and test if this works fine for your S922X too?
@Ntemis can you check if it builds fine for RK3288?

Builds fine for all platforms & was tested on a RockPro64.

I was wondering if we could change this [elif](https://github.com/libretro/mupen64plus-libretro-nx/pull/74/files#diff-2e18049090f9b3589548e63b7f310e21R213) to else to always fall back on `/usr/lib/libGLESv2.so` if nothing else is defined & OpenGL ES should be used.

### DIFF
--- a/GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
+++ b/GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
@@ -7,13 +7,9 @@
 #include <winlnxdefs.h>
 #endif
 
-#ifdef UNDEF_GL_GLEXT_PROTOTYPES
-// Issues on Raspberry since GL2.h defines it and causes
-// Macro shenanigans
 #ifdef GL_GLEXT_PROTOTYPES
 #undef GL_GLEXT_PROTOTYPES
 #endif // GL_GLEXT_PROTOTYPES
-#endif // UNDEF_GL_GLEXT_PROTOTYPES
 
 #ifdef EGL
 #include <GL/glcorearb.h>
--- a/GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
@@ -12,7 +12,7 @@
 #define glGetProcAddress wglGetProcAddress
 #define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) glGetProcAddress("gl"#proc_name)
 
-#elif defined(VERO4K) || defined(ODROID) || defined(VC) || defined(AMLOGIC)
+#elif defined(VERO4K) || defined(ODROID) || defined(VC) || defined(USE_GENERIC_GLESV2)
 
 #include <dlfcn.h>
 #define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) dlsym(gles2so, "gl"#proc_name);
@@ -210,7 +210,7 @@ extern "C" void initGLFunctions()
 	void *gles2so = dlopen("/usr/lib/arm-linux-gnueabihf/libGLESv2.so", RTLD_NOW);
 #elif defined(VERO4K)
        void *gles2so = dlopen("/opt/vero3/lib/libGLESv2.so", RTLD_NOW);
-#elif defined(AMLOGIC)
+#elif defined(USE_GENERIC_GLESV2)
        void *gles2so = dlopen("/usr/lib/libGLESv2.so", RTLD_NOW);
 #endif
 
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ else ifneq (,$(findstring rpi,$(platform)))
       CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard
       HAVE_NEON = 1
    endif
-   COREFLAGS += -DUNDEF_GL_GLEXT_PROTOTYPES -DOS_LINUX
+   COREFLAGS += -DOS_LINUX
    ASFLAGS = -f elf -d ELF_TYPE
 
 # Nintendo Switch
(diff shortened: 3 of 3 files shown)


shantigilbert:
I will check tonight and will let you know ASAP.

5schatten:
You probably have to adjust your Mupen64Plus-NX package a bit to use the correct platform for the specific SoC :smiley:

Ntemis:
Well done your changes check out.
GTG on my part.

shantigilbert:
Compiles fine :)

5schatten:
Sure no problem so drop `UNDEF_GL_GLEXT_PROTOTYPES` from COREFLAGS? Anything else to clean up?

m4xw:
@5schatten Drop it from all the makefile entries, afterwards `GLideN64\src\Graphics\OpenGLContext\GLFunctions.h` +10-16, remove the `#ifdef UNDEF_GL_GLEXT_PROTOTYPES/#endif` around the `#ifdef GL_GLEXT_PROTOTYPES` block

m4xw:
And I guess you can change the comment

barbudreadmon:
same comment as https://github.com/libretro/yabause/pull/90 : there are distros using those names already, changing rockpro64 to RK3399 will break their build scripts, is this necessary ?

GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
```diff
@@ -7,14 +7,6 @@
 #include <winlnxdefs.h>
 #endif
 
-#ifdef UNDEF_GL_GLEXT_PROTOTYPES
```
m4xw: Don't fully remove it.
Keep the 
```
#ifdef GL_GLEXT_PROTOTYPES
#undef GL_GLEXT_PROTOTYPES
#endif // GL_GLEXT_PROTOTYPES
```

5schatten:
> same comment as [libretro/yabause#90](https://github.com/libretro/yabause/pull/90) : there are distros using those names already, changing rockpro64 to RK3399 will break their build scripts, is this necessary ?

As I wrote in the Yabasanshiro PR the platform should be SoC specific if possible & "Amlogic" does produce plenty of them. As @shantigilbert added the Amlogic platform in the first place I guess it's fine to make some adjustments to have several SoCs covered.

shantigilbert:
From my point of view this PR makes sense, as it will cover many more devices. But maye to fix this issue with other distros do a check for rockpro64 then just replace the platform with proper platform ? 

```
ifneq (,$(findstring rockpro64,$(platform)))
      platform = RK3399
```

5schatten:
There was no rockpro64 platform before so if a distro want to use Rockchip specific flags they can start with RK3399 too. I guess it pointed to your amlogic platform which I added back.

shantigilbert:
Ah I was basing my comment from what @barbudreadmon wrote on the Yabasanshiro pr, sorry.

GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
5schatten: Should be fine now?

5schatten:
@shantigilbert no problem - I'll add these SoC platforms and keep the board platforms for now. I could imagine that over time maintainers will adopt their scripts to use them because it's more flexible.

Anyway a common scheme for all libretro cores would be useful. Maybe also something like `$Vendor-$Userspace-$SoC` would be clever. So Amlogic could define some common stuff, then then 32 or 64 adds the usual arm or aarch64 & dynarec stuff and the you'll set SoC flags afterwards. :thinking: 

But I guess it should be fine the way it is right now.

m4xw:
@5schatten Is this ready to merge?
Also any specific reason using the arm dynarec on rockpro64?
No aarch64 kernel?
(Simliar question re: amlogic)

5schatten:
The kernel is `arch=aarch64` but the userspace is `arch=arm` so I assume I shouldn't mix that? I use these cores for my LibreELEC fork and LE builds a 64bit kernel but 32bit userspace which is needed for Netflix for example.

m4xw:
Well yea, you can't link them together, however if you have a aarch64 RA (or simliar) build, it will work.
armv7 is currently in the process of being deprecated by vendors etc, so aarch64 is recommended going forward.
I will probably merge it later or tomorrow.

Ntemis:
More simple reason.
Most current cores build as arch=aarch64 perform poorly as they lack a dynarec for arm64.

m4xw:
@Ntemis I am aware, however most emulators these day support or work on aarch64 support.
The only real exception being PSX.
It's basically a downgrade across the board because single emulators don't support it.
Also I had to deal with this exact issue myself, hence we got the aarch64 dynarec bounty going instead of trying to get aarch32 execution mode to work for Horizon.

Ntemis:
Is not something that's going to be fixed overnight, so merge this and we keep a hopefully thinking

m4xw:
Oh its def. not smth that's fixed over night.
Don't get me wrong^^

5schatten:
Correct me if I'm wrong but it's still armv8-a but only running userland in 32bit mode? Afaik aarch32 can make use of all armv8-a instruction so it's not entirely the same as armv7-a. Also you need a OS that is completely compiled for aarch64 and afaik Raspbian is still using 32bit same as Lakka or LibreELEC. And the latter is probably one of the "bigger" distros for SBCs so I guess if they make the switch to 64bit others will follow.

m4xw:
aarch32 is using a different ISA that's backwards compatible with armv7.
You would also need to ship 32 + 64bit user-space libraries for this to work as well as have different frontend builds unless you want to do some funky EL0 stuff.
About the address limitations, these hardly matter IMO.
The main downside is you are limited to 32bit registers in aarch32, which will always be inferior if optimized properly for aarch64.
Also re: PCSX2, good that Play! is a thing (includes a aarch64 dynarec!).

m4xw:
Thanks for the PR btw.

5schatten:
Yes of course is aarch32 different to aarch64 but it can make use of at least some armv8-a stuff to. Well afaik :see_no_evil: 
Mixing 32&64 bit works "fine" for other distros like Arch & multilib but that makes no sense for me for SBC purposes where I can compile the whole distro by myself & tailor it for my needs. IIRC all major emulators feature now a dynarec for 64 too so this should work fine but you have to consider other purposes as well. As mentioned before LibreELEC offers support for Netflix *but* this only works for 32bit systems, for whatever reasons. So if you consider all costs and benefits I guess using 64bit while sacrificing other features isn't worth it for now.

Beside that I probably can rework the platform opts later to consider 32bit as well but I can't test the 64bit build at all.

m4xw:
I am not complaining, keeping compat. is very important^^

5schatten:
Yes of course. We're not Apple after all :laughing:

m4xw:
Or Google :thonk:

PR #82, opened 2019-08-04
Fix Raspberry Pi 4 build configuration

jdonald:
This branch cherry-picks ematysek's last commit from the (recently locked) [mupen64plus-libretro repo](https://github.com/libretro/mupen64plus-libretro).

I then fixed it so things just work as expected on Buster with:

    make platform=rpi4 -j4

This will link with Mesa GLES because the legacy brcmGLES driver is not functional on a Pi 4. Also fixed a null pointer crash on launch which may be specific to the version of Mesa used on Buster.

### DIFF
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -54,8 +54,8 @@ bool Utils::isEGLExtensionSupported(const char * extension)
 	EGLDisplay display = eglGetDisplay(EGL_DEFAULT_DISPLAY);
 	const char* extensions = eglQueryString(display, EGL_EXTENSIONS);
 	if (extensions == nullptr) {
-		EGLDisplay platformDisplay = eglGetPlatformDisplay(EGL_PLATFORM_GBM_KHR, display, nullptr);
-		extensions = eglQueryString(platformDisplay, EGL_EXTENSIONS);
+		// FIXME: KMS/DRM requires EGL_PLATFORM_GBM_KHR and libgbm, see mupen64plus-libretro-nx#82
+		return true;
 	}
 
 	const char* start = extensions;
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -51,10 +51,15 @@ bool Utils::isEGLExtensionSupported(const char * extension)
 	if (where || *extension == '\0')
 		return false;
 
-	const char* extensions = eglQueryString(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_EXTENSIONS);
+	EGLDisplay display = eglGetDisplay(EGL_DEFAULT_DISPLAY);
+	const char* extensions = eglQueryString(display, EGL_EXTENSIONS);
+	if (extensions == nullptr) {
+		EGLDisplay platformDisplay = eglGetPlatformDisplay(EGL_PLATFORM_GBM_KHR, display, NULL);
+		extensions = eglQueryString(platformDisplay, EGL_EXTENSIONS);
+	}
 
 	const char* start = extensions;
-	while (start) {
+	for (;;) {
 		where = strstr(start, extension);
 		if (where == nullptr)
 			break;
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -51,8 +51,7 @@ bool Utils::isEGLExtensionSupported(const char * extension)
 	if (where || *extension == '\0')
 		return false;
 
-	EGLDisplay display = eglGetDisplay(EGL_DEFAULT_DISPLAY);
-	const char* extensions = eglQueryString(display, EGL_EXTENSIONS);
+	const char* extensions = eglQueryString(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_EXTENSIONS);
 	if (extensions == nullptr) {
 		// FIXME: KMS/DRM requires EGL_PLATFORM_GBM_KHR and libgbm, see mupen64plus-libretro-nx#82
 		return true;
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -54,7 +54,7 @@ bool Utils::isEGLExtensionSupported(const char * extension)
 	const char* extensions = eglQueryString(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_EXTENSIONS);
 
 	const char* start = extensions;
-	for (;;) {
+	while (start) {
 		where = strstr(start, extension);
 		if (where == nullptr)
 			break;
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -54,7 +54,7 @@ bool Utils::isEGLExtensionSupported(const char * extension)
 	EGLDisplay display = eglGetDisplay(EGL_DEFAULT_DISPLAY);
 	const char* extensions = eglQueryString(display, EGL_EXTENSIONS);
 	if (extensions == nullptr) {
-		EGLDisplay platformDisplay = eglGetPlatformDisplay(EGL_PLATFORM_GBM_KHR, display, NULL);
+		EGLDisplay platformDisplay = eglGetPlatformDisplay(EGL_PLATFORM_GBM_KHR, display, nullptr);
 		extensions = eglQueryString(platformDisplay, EGL_EXTENSIONS);
 	}
 
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,11 @@ ifneq (,$(findstring unix,$(platform)))
 else ifneq (,$(findstring rpi,$(platform)))
    TARGET := $(TARGET_NAME)_libretro.so
    LDFLAGS += -shared -Wl,--version-script=$(LIBRETRO_DIR)/link.T -Wl,--no-undefined -ldl
-   GLES = 1
+   ifeq ($(FORCE_GLES3),1)
+      GLES3 = 1
+   else
+      GLES = 1
+   endif
    ifneq (,$(findstring mesa,$(platform)))
       MESA = 1
    endif
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,9 @@ else ifneq (,$(findstring rpi,$(platform)))
    else ifneq (,$(findstring rpi3,$(platform)))
       CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard
       HAVE_NEON = 1
+   else ifneq (,$(findstring rpi4,$(platform)))
+      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard
+      HAVE_NEON = 1
    endif
    COREFLAGS += -DOS_LINUX
    ASFLAGS = -f elf -d ELF_TYPE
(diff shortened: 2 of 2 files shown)


m4xw:
You should ensure it picks at least GLESv3, since it's free real estate for Pi4.
Also I am sure the EGL change will break a few things, not sure what tho.
Other did simliar changes, @natinusala would you try on hw?

m4xw:
cc @hhromic

m4xw:
Also with GLES3 i'd make sure "VC" isn't defined, since then theres no reason for the RPi specific stuff.

hhromic:
Thanks for the heads-up @m4xw .

I also don't like very much the solution to the EGLQuery problem here. In RetroPie we tried that same approach first, with no apparent problems. However after examining the actual code-path of the affected function we ended-up using a different solution (See [1]).

We verified that the EGL extensions being queried are actually supported so we force returning `true` for RPI4 instead of `false` as in this PR when the pointer is NULL.

Of course our solution is also not ideal, but we can't figure out yet why the EGLQuery function returns NULL. Perhaps as the OP of this PR mentions, is a mesa issue.

[1] https://github.com/RetroPie/RetroPie-Setup/blob/master/scriptmodules/libretrocores/lr-mupen64plus-next/0001-force-egl.patch

natinusala:
GLES3 should be enabled in the rpi4 platform IMO

Sadly I can't test anytime soon as I'm away from home.

hhromic:
Ah, also worth to note that the RPI3 can also run in fkms/drm mode with mesa drivers in Buster, and actually seems to work better than the legacy drivers (more game testing is still needed). However, only GLES2 is supported there.

So in summary, there should be a `platform=rpi3` (legacy), `platform=rpi3-mesa` (mesa and GLES2) and `platform=rpi4` (mesa and GLES3).

jdonald:
> We verified that the EGL extensions being queried are actually supported

Thanks for the heads-up. I'm coming up with a cleaner change here for sake of clarify and sake of documenting to file a ticket on Mesa . However, so far I am unable to test it. If you look at both uses of `m_glinfo.eglImage` they are gated off with with `#ifdef OS_ANDROID`. @hhromic how did you verify that `EGL_KHR_image` is working as expected on Raspbian Buster?

> GLES3 should be enabled in the rpi4 platform IMO

This is not a difficult change to build and test quickly, but hard to make it an elegant pull request. You need to either:
* Change many lines across many source files, including for example the many instances of `if (_glinfo.isGLES2)` to `if (_glinfo.isGLES2 || _glinfo.isGLES3)`, in addition to the Makefile logic, and take care to get all the || and &&s right; or
* Change a smaller set of code that sets up these variables, but document that `isGLESX` means 1, 2, or 3, while `isGLES2` actually means 2 or 3, which would make the code harder to reason about going forward.

I recommend we keep this pull request on hand targeting only GLES2 in hopes of making the master branch functional first, then have a later pull request on top of that with the more involved review with changes across a dozen+ sources files.

m4xw:
@jdonald GLES3 is already completely functional. GLES3=1 and FORCE_GLES3=1

jdonald:
@m4xw while those variables exist in the Makefile, this is all it appears to do:
```
else ifeq ($(GLES3),1)
        GLFLAGS += -DEGL -DHAVE_OPENGLES -DHAVE_OPENGLES3 -DGLES3
        LDFLAGS += $(EGL_LIB)
        SOURCES_C += $(LIBRETRO_COMM_DIR)/glsym/glsym_es3.c
        ifeq ($(ANDROID),1)
                SOURCES_CXX += $(VIDEODIR_GLIDEN64)/src/Graphics/OpenGLContext/GraphicBuffer/GraphicBufferWrapper.cpp \
                               $(VIDEODIR_GLIDEN64)/src/Graphics/OpenGLContext/GraphicBuffer/PublicApi/android_hardware_buffer_compat.cpp
        endif
```
glsym_es3.c just lists out the symbols and is practically the same as glsym_es2.c

I was going by the fact that there are zero lines in the code with `#include <GLES3/...>`. I guess I can indeed put this in the Makefile and if it was a no-op before it will continue to work. On the bright side, that means a plugin compiled for rpi4 will still work on a Pi 3 because if in reality it's only using GLES2 features.

I do spot one bug in `opengl_GLInfo.cpp` for non-forced GLES3 but that should be simple to fix (any may also explain why people have unexpected performance issues on the Pi 4).

natinusala:
Because gles3 includes do not exist, nor does a gles3 .so

m4xw:
It uses generated header files from GLSM

jdonald:
@natinusala header files can be found at e.g. `/usr/include/GLES3/gl3ext.h` and are installed as part of `libgles2-mesa-dev`. Correct that both GLES2 and GLES3 symbols are in `libGLESv2.so` though.

@m4xw I have built with GLES 3 now, but looking through glsm.c and a lot of the code looks suspicious, lines like `#ifndef HAVE_OPENGLES2` that should say `#if !defined(HAVE_OPENGLES2) && !defined(HAVE_OPENGLES3)`. This is in at least 10 places where apparently it goes to the GLES1 codepath for the `-DHAVE_OPENGLES3 -DGLES3` mdoe of the build. You said GLES3 works with GlideN64 has been fully tested? If x86_64 Linux, I suspect that serious performance problems could have gone unnoticed.

m4xw:
GLES3 always defines GLES2 and yes, it should be fully functional.

jdonald:
> GLES3 always defines GLES2

I just did a very quick test to show this is not the case in glsm.h:
```
 #if defined(HAVE_OPENGLES3)
  #ifndef HAVE_OPENGLES2
    #error defined HAVE_OPENGLES3 but not HAVE_OPENGLES2!!
  #endif
#endif
```
The current Makefile GLES3 logic (used on RK3399 and AMLG12) sets the two separately.

If GLES3 was fully tested for functionality on RK3399 and other ARM devices with GLES3, my command still stands that there may be performance issues for the reason above.

m4xw:
I will investigate

jdonald:
By the way, with the current Makefile logic `make platform=rpi3 FORCE_GLES3=1` is currently a no-op for anything other than `platform=unix`, and make `make platform=rpi3 GLES3=1` also has no effect due to the GLES check coming first deep in Makefile.common.

That might explain if things to appear functional when testing this on unsupported devices. RK3399 and AMG12x detect it conditionally so it might work there. However, there's still the obvious bug in dynamic detection (string check for "OpenGL ES 2" in `opengl_GLinfo.cpp` and perhaps that combined mismatch makes things work (in GLES 2.0 mode).

hhromic:
> @hhromic how did you verify that EGL_KHR_image is working as expected on Raspbian Buster?

I found a test program that prints the EGL extensions using mesa with `eglQueryString()` (see instructions below). This example WORKS with current mesa/fkms/drm in Buster, which makes me unsure if Mesa is at fault here.

I instrumentalised the call in mupen64plus-next and it seems to be getting a BAD_DISPLAY error after the call (hence why it returns NULL in the first place). I have a hunch that RetroArch is not giving access to the current display and then `eglQueryString()` can't query it as expected. But no idea why.

---

Quick instructions to get the supported EGL extensions from Mesa in Buster:
```
$ curl -SLO https://gist.githubusercontent.com/Miouyouyou/89e9fe56a2c59bce7d4a18a858f389ef/raw/b4b5c03398ef84446f9212e1112d87dcaee8fa87/Linux_DRM_OpenGLES.c
$ gcc -o drmgl Linux_DRM_OpenGLES.c `pkg-config --cflags --libs libdrm` -lgbm -lEGL -lGLESv2
$ ./drmgl
Using display 0x125dfd8 with EGL version 1.4
EGL Version "1.4"
EGL Vendor "Mesa Project"
EGL Extensions "EGL_ANDROID_native_fence_sync EGL_EXT_buffer_age EGL_EXT_image_dma_buf_import EGL_EXT_image_dma_buf_import_modifiers EGL_KHR_cl_event2 EGL_KHR_config_attribs EGL_KHR_create_context EGL_KHR_create_context_no_error EGL_KHR_fence_sync EGL_KHR_get_all_proc_addresses EGL_KHR_gl_colorspace EGL_KHR_gl_renderbuffer_image EGL_KHR_gl_texture_2D_image EGL_KHR_gl_texture_3D_image EGL_KHR_gl_texture_cubemap_image EGL_KHR_image EGL_KHR_image_base EGL_KHR_image_pixmap EGL_KHR_no_config_context EGL_KHR_reusable_sync EGL_KHR_surfaceless_context EGL_EXT_pixel_format_float EGL_KHR_wait_sync EGL_MESA_configless_context EGL_MESA_drm_image EGL_MESA_image_dma_buf_export EGL_WL_bind_wayland_display "
GL Extensions: "GL_EXT_blend_minmax GL_EXT_multi_draw_arrays GL_EXT_texture_format_BGRA8888 GL_OES_compressed_ETC1_RGB8_texture GL_OES_depth24 GL_OES_element_index_uint GL_OES_fbo_render_mipmap GL_OES_mapbuffer GL_OES_rgb8_rgba8 GL_OES_stencil8 GL_OES_texture_3D GL_OES_texture_npot GL_OES_vertex_half_float GL_OES_EGL_image GL_OES_depth_texture GL_AMD_performance_monitor GL_OES_packed_depth_stencil GL_OES_get_program_binary GL_APPLE_texture_max_level GL_EXT_discard_framebuffer GL_EXT_read_format_bgra GL_EXT_frag_depth GL_NV_fbo_color_attachments GL_OES_EGL_image_external GL_OES_EGL_sync GL_OES_vertex_array_object GL_EXT_occlusion_query_boolean GL_EXT_unpack_subimage GL_NV_draw_buffers GL_NV_read_buffer GL_NV_read_depth GL_NV_read_depth_stencil GL_NV_read_stencil GL_EXT_draw_buffers GL_EXT_map_buffer_range GL_KHR_debug GL_KHR_texture_compression_astc_ldr GL_OES_required_internalformat GL_OES_surfaceless_context GL_EXT_separate_shader_objects GL_EXT_compressed_ETC1_RGB8_sub_texture GL_EXT_draw_elements_base_vertex GL_EXT_texture_border_clamp GL_KHR_context_flush_control GL_OES_draw_elements_base_vertex GL_OES_texture_border_clamp GL_KHR_no_error GL_KHR_texture_compression_astc_sliced_3d GL_KHR_parallel_shader_compile GL_MESA_tile_raster_order "
```

You can notice that `EGL_KHR_image_base` and `EGL_KHR_image` are both reported as supported EGL extensions.

hhromic:
I just got confirmation from another RetroPie dev that the `eglQueryString()` returning NULL issue also happens in standalone GLideN64 at least in KMS/DRM mode in Buster.

This implies that RetroArch/mupen-next core are highly likely not at fault indeed.

jdonald:
Thanks @hhromic I looked at the implementation in `Linux_DRM_OpenGLES.c` and I think the proper fix would be to use `EGL_PLATFORM_GBM_KHR`. However, the changes would snowball if I included headers and linked with `libgbm.so`.

Note my earlier comment: whether we return false or true for `EGL_KHR_image_base` only has a functional difference in the case of `OS_ANDROID`. KMS support for Android is still in its infancy, and with the hack perhaps we can bank on those bleeding edge Android Q builds with KMS also supporting `EGL_KHR_image` for now.

I have also made it so this branch now supports builds with both:

    make platform=rpi4

and

    make platform=rpi4 FORCE_GLES3=1

That should make it straightforward here for others to test, particularly for questions above around the `HAVE_OPENGLES3` codepath.

hhromic:
> I looked at the implementation in Linux_DRM_OpenGLES.c and I think the proper fix would be to use EGL_PLATFORM_GBM_KHR. However, the changes would snowball if I included headers and linked with libgbm.so.

That's very nice you found the actual culprit and proper solution. I agree with your approach too. I saw in RetroArch the same is done with `EGL_PLATFORM_GBM_KHR` to get the display. Too bad it requires yet another dependency (libdrm, which RetroArch actually does require) :/

> Note my earlier comment: whether we return false or true for EGL_KHR_image_base only has a functional difference in the case of OS_ANDROID. KMS support for Android is still in its infancy, and with the hack perhaps we can bank on those bleeding edge Android Q builds with KMS also supporting EGL_KHR_image for now.

I took a closer look to what you said and indeed you are completely right that it has no effect for non-Android builds. So for RPI it doesn't matter what we return there :)

Thanks for your PR, I see it enables all the following:

    make platform=rpi3  # to use old legacy driver for RPI3
    make platform=rpi3-mesa  # to use new fkms/drm driver in RPI3
    make platform=rpi4  # to use new fkms/drm driver in RPI4

I tested `make platform=rpi3-mesa` and is building/working just fine (except for some graphical glitches with depth that are not really a problem of this PR).

m4xw:
Oh PR got closed when I changed branches.

m4xw:
@jdonald Can you retarget to develop please

jdonald:
> Oh PR got closed when I nuked branches.

Understandable. Filed new PR #95.

PR #92, opened 2019-09-04
Add configs for Classic consoles

evilwraith:


m4xw:
There's also a third solution, `while(start)`
If Init is missing it def needs fixing upstream, but not sure if this is a core issue as I think the frontend should handle that?
We also should maybe ping @gonetz on that one

m4xw:
fwiw I have seen these nullptr's on Raspberry and odroid only.

hhromic:
That problem has been there for a while now and heavily discussed in #82 .
And indeed it manifests in RPI builds with FKMS+Mesa.

Imho @jdonald found the proper explanation for this bug here (the need to use `EGL_PLATFORM_GBM_KHR` because RetroArch uses GBM with KMS): https://github.com/libretro/mupen64plus-libretro-nx/pull/82#issuecomment-518355169

Moreover, in his same comment he also pointed out that querying for egl extensions is only relevant for Android, which seems to not return `nullptr` (so all good), therefore the check for `nullptr` should be a safe workaround.

But yes, I also agree @gonetz's opionion is key here.

gonetz:
I'm afraid I'm not competent in this matter. Egl initialization for GLideN64 was implemented by @fzurita. If eglInitialize call is required for Raspberry and odroid only, you may make it platform-specific, like:
```
#ifdef ODROID
eglInitialize(eglGetDisplay(EGL_DEFAULT_DISPLAY), nullptr, nullptr);
#endif

```

m4xw:
@hhromic can you check for any issues with the the broadcom drivers if eglInitialize is used?

hhromic:
Yes I was just going to comment that this `nullptr` issue only manifests when using KMS in Raspberry Pi, not when using the legacy broadcom drivers. So in summary, when RetroArch is running in KMS+GBM+DRM mode in Raspberry Pi we need to use `EGL_PLATFORM_GBM_KHR` with `eglGetPlatformDisplay()` instead of `EGL_DEFAULT_DISPLAY` with `eglGetDisplay()`. Also, this requires `libgbm` which is already a dependency of RetroArch.

Standalone mupen64plus using KMS in Raspberry Pi doesn't seem to need this (not sure if it uses `libgbm`), but I'm going to confirm during the day with the RetroPie developer who was testing standalone with KMS.

m4xw:
@hhromic If it doesn't cause any side-effects i'd say it's whatever.
If you have a better proposal, I am all ears.
Feels to be tho this should be called in RA, not so sure.

hhromic:
I can confirm now from the other RetroPie dev that:
```cpp
const char* extensions = eglQueryString(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_EXTENSIONS);
```
also returns `nullptr` in **standalone** Mupen when using KMS/DRM in Raspberry Pi, so is not a libretro-mupen-next exclusive issue.

The current dirty patch in RetroPie simply does this in Raspberry Pi with KMS/DRM:
```diff
-	eglImage = (Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
+	eglImage = true; //(Utils::isEGLExtensionSupported("EGL_KHR_image_base") || Utils::isEGLExtensionSupported("EGL_KHR_image"));
```
Which avoids completely calling `isEGLExtensionSupported()`. There are no observed side-effects, specially because these extensions are only used for Android devices (nothing else checks for `eglImage`).

However, I don't like that patch and I would take instead the suggestion given by @gonetz (gating eglInit behind a compile-define). I can test this solution on RPI soon. I also think that given that @fzurita is the original implementor in upstream, we should wait for him to chime-in and hopefully fix this upstream.

hhromic:
> Feels to me tho this should be called in RA, not so sure.

The solution from @jdonald is more in-line to that as it gets the display from the platform initialised by RA via GBM. Afaik there is no API for libretro cores to ask for the display to RA , is it?

Anyway I checked RA and the good news are that if we need `EGL_PLATFORM_GBM_KHR`, we can simply define it ourselves instead of depending on `libgbm` like RetroArch does itself:
```
gfx/drivers_context/drm_ctx.c:#ifndef EGL_PLATFORM_GBM_KHR
gfx/drivers_context/drm_ctx.c:#define EGL_PLATFORM_GBM_KHR 0x31D7
```
However I'm not sure this solution aligns with upstream :/
As I said before, let's wait for the original implementor to chime-in.

m4xw:
Yea agree

jdonald:
I don't have my Pi at the moment. Have you guys confirmed that adding the eglInitialize() line gets eglQueryString() to return something reasonable in KMS/DRM? If so, it's a superior solution to the two workarounds. If not, it's worse.

> Anyway I checked RA and the good news are that if we need EGL_PLATFORM_GBM_KHR, we can simply define it ourselves

I thought the problem is that we need a device context from libgbm to pass e.g.:

    get_platform_display(EGL_PLATFORM_GBM_KHR, gbm.dev, NULL);

so providing the #define alone may not be sufficient. It's been a while, but I recall it didn't work if I tried passing nullptr for that second arg.

Unrelated bits I noticed about this PR:

1.

   $(shell echo `$(CC) -dumpversion` ">= 5" | bc -l)

always returns a parse error in my testing, because `bc` expects numbers to have at most one decimal point and thus can't compare 8.3.0 >= 5. At least it means this code won't ever statically link libc by surprise.

2.

Apparently one configuration uses `-fno-rtti` and another doesn't. If this is a no-RTTI codebase it should be so consistently to remove that overhead. However, it appears that a limited number of `throw` statements are conditional under configs like GL_DEBUG, leading to a mess.

3.

`-mcpu` is deprecated on x86 and likely on the way out for ARM/AArch64, while the equivalent -march + -mtune is preferred. Looks like this setup oddly uses -march for classic_armv7 but not classic_armv8_a35.

evilwraith:
@James Donald I've been slowly cleaning up our old configs. Guess I missed
that in the PSClassic one. Will check on that.

I can confirm we've been having this problem with the MUPEN64-based cores
on the PSC for some time now, but weren't sure what fix would be
acceptable.

hhromic:
> I thought the problem is that we need a device context from libgbm to pass e.g.:
> get_platform_display(EGL_PLATFORM_GBM_KHR, gbm.dev, NULL);
> so providing the #define alone may not be sufficient. It's been a while, but I recall it didn't work if I tried passing nullptr for that second arg.

Umm, from your [commit](https://github.com/libretro/mupen64plus-libretro-nx/pull/82/commits/1c75fe454ed8756720b99a5b841da95d8298267d) it doesn't seem to use anything but `EGL_PLATFORM_GBM_KHR`.

Anyway I tried that fix now on my RPI3 with Buster and KMS and it **doesn't work** unfortunately. 
I tried the solution given in this PR of adding this:

    eglInitialize(eglGetDisplay(EGL_DEFAULT_DISPLAY), nullptr, nullptr);

and it also **doesn't work** on the RPI3 with Buster and KMS unfortunately.

What **does work** is simply returning `false` when `extensions == nullptr` or the equivalent solution `while (start)` that also returns `false` when `extensions == nullptr`.

@m4xw given the fact that all pieces of code that use the result of `isExtensionSupported()` are gated behind an `OS_ANDROID` define (I manually verified this just now), I would safely assume that just returning `false` when we can't obtain the EGL extensions is the better solution for now.
No side-effects seen so far anyway during testing now and before.

I would even upstream it as it's a very harmless/normal safety guard against a possible null pointer.

m4xw:
What a drag, lol

hhromic:
Can't agree more :P

So this is my proposed change after all the fuzz (the first check is just in case and for consistency, it doesn't actually trigger in my setup):
```diff
diff --git a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
index 0279af9..a7bd45e 100644
--- a/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/opengl_Utils.cpp
@@ -27,6 +27,10 @@ bool Utils::isExtensionSupported(const opengl::GLInfo & _glinfo, const char *ext
                return false;

        const GLubyte *extensions = glGetString(GL_EXTENSIONS);
+       if (extensions == nullptr) {
+               LOG(LOG_WARNING, "[GlideN64]: could not query GL extensions on this device");
+               return false;
+       }

        const GLubyte *start = extensions;
        for (;;) {
@@ -52,6 +56,10 @@ bool Utils::isEGLExtensionSupported(const char * extension)
                return false;

        const char* extensions = eglQueryString(eglGetDisplay(EGL_DEFAULT_DISPLAY), EGL_EXTENSIONS);
+       if (extensions == nullptr) {
+               LOG(LOG_WARNING, "[GlideN64]: could not query EGL extensions on this device");
+               return false;
+       }

        const char* start = extensions;
        for (;;) {
```
If you like it we can either send it upstream or make a separate PR and call it a day for good.

jdonald:
> from your [commit](https://github.com/libretro/mupen64plus-libretro-nx/pull/82/commits/1c75fe454ed8756720b99a5b841da95d8298267d) it doesn't seem to use anything but `EGL_PLATFORM_GBM_KHR`.
> 
> Anyway I tried that fix now on my RPI3 with Buster and KMS and it **doesn't work** 

Right. Sorry, that commit was a work-in-progress. At first I thought it was working properly, but quickly realized at that time I had not tested KMS. That's why the commits following it in #82 remove that code and resort to only checking `extensions == nullptr`.

As you have confirmed that `eglInitialize()` does not solve the problem either, I agree that checking nullptr is the right solution.

However, while you return `false`, in #82 I erred on the side of returning `true` to make this work in theory for KMS/DRM on Android. You mentioned that Android is not returning nullptr here, but it's a bit unclear if you tested Android with KMS/DRM, which I'm not even sure how to do.

hhromic:
> However, while you return false, in #82 I erred on the side of returning true to make this work in theory for KMS/DRM on Android. You mentioned that Android is not returning nullptr here, but it's a bit unclear if you tested Android with KMS/DRM, which I'm not even sure how to do.

No I did not test on Android, only on RPI with KMS/DRM and RPI with Broadcom drivers.

I mentioned that on Android it does not return nullptr because otherwise it would be segfaulting there as well and nobody has reported that as far as I know. I assume plenty of people use this core on Android too. Therefore I concluded that it works properly there. Also, on RPI with Broadcom legacy drivers it also doesn't segfault, meaning that it also does not return nullptr in that setup.

jdonald:
> I assume plenty of people use this core on Android too

KMS/DRM on Android is so bleeding edge that there probably aren't any consumer phones using it yet, but that may change in the future. It certainly seems fine to punt on this until when/if it ever becomes a problem. The LOG_WARNING will be there when the time comes to remind a dev what's happening (rather than silently dropping performance due to not using `EGL_KHR_image`).

> Also, on RPI with Broadcom legacy drivers it also doesn't segfault, meaning that it also does not return nullptr in that setup.

Yeah, the brcmGLES examples under `/opt/vc/src/hello_pi` all just use `eglGetDisplay(EGL_DEFAULT_DISPLAY)` and they work, so this should too.

fzurita:
```eglInitialize``` is a required call. In android, we are doing it here outside of the video plugin:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/mupen64plus-ui-console/ae-bridge/ae_vidext.cpp#L99

I would suggest that the mupen64plus vidext API should also be implemented in mupen64plus-libretro-nx, but I'm not familiar enough with libretro to know how it would be done there.

Also EGL Image is only used by Android, and Android support of very iffy. Adreno devices support it very well with the public API in Android API level 26 and greater. Older devices with Android API 23 and older also support it somewhat well (Mali, Adreno, and PowerVR).

EGL Image is currently used to perform copies from video memory to system memory. It makes copies nearly instant with no performance penalty unlike ```glReadPixels```. In the future, it could also be used to much more quickly copy textures to video memory as well.

hhromic:
@fzurita thanks for the clarification, much appreciated.

Currently there is no call to `eglInitialize` anywhere in mupen64plus-libretro-nx, so for the moment I guess it doesn't seem to be a priority and seems to be a bigger fish to fry.

However, what you explained makes very clear to me that checking for eglImage is only relevant for Android and the proposed fix of just returning `false` when we can't query EGL extensions is 100% safe and the suitable fix for the problem at hand. Thanks again for the insights.

@m4xw would you like to go ahead with this solution so we can close this problem for good?  In my opinion it should upstreamed but better you tell what is the best way to proceed.

m4xw:
@fzurita according to you, in libretro, it would be a frontend issue.
Do you know if there are issues with multiple initializations?
If not might as well just do it in gliden, but also add a check for extensions (see lower).

@hhromic I've actually inspected some enterprise projects like Qt (also am a Qt engineer^^) and it also does a sanity check for the egl extension return, while also doing eglInitialize (also for kms stuff).

Offtopic: I've also played around with it and made updated builds with new toolchains and crosscompile envs that i might publish soon for Pi4, but awful lots of driver issues. Gotta say so far not a good experience. Also broadcom drivers seem to have issues with more advanced applications, mesa is plagued by severe tearing issues.
I might gonna do some bugfixes for mesa on company time if its worth.
For some reason its also a lot more pronounced on aarch64 ubuntu than aarch32 raspbian.

hhromic:
@m4xw we have been investigating the severe tearing issue in Mesa/KMS/DRM with the RetroPie folks and the conclusion is that when the output frame rate is way below the display frequency the tearing goes bananas. We found that activing the frame duping and fullspeed framerate options completely eliminates the tearing on RPI + Mesa/KMS/DRM because the emulator provides framerates much closer to the display refresh (vsync).  Also the RPI foundation made available a release candidate of Mesa 19.2.0 that fixes a number of issues.

I myself have been testing a number of games with this setup: RPI3B+/Plain Official Raspbian Buster aarch32/Mesa 19.2 RC/KMS/DRM + frameduping + fullspeed framerate  and the results are very satisfactory for me. Much less glitches than with the legacy broadcom drivers.

So what's the conclusion about the nullptr check? Do you want me to send a PR or you want to do it yourself? I would like to fix that so we can avoid manual local patches :) :)

m4xw:
@hhromic let's talk about how to continue tomorrow on discord.
as for the post I was actually referring to a qt application I have been building.
I think there is more going on, I literally manage to kill the rpi Soc reliable via gpu stuff

m4xw:
Also please include aarch64 testing, I thing this is very much needed and neglected by everyone. I will keep a keen eye on it myself and will try to optimize for it.

jdonald:
Hi guys thanks for pushing to resolve the KMS crash.

> For some reason its [tearing] also a lot more pronounced on aarch64 ubuntu than aarch32 raspbian.

Sanity check: in glxinfo’s output what version of Mesa are you running on Ubuntu and is it successfully getting the driver for Broadcom V3D + GLES 3.0?

The V3D fix for 64-bit kernels got merged two weeks ago and half of the Pi4 aarch64 distros still lack it (i.e. softpipe rendering). Furthermore, afaik Ubuntu 19.04 landed with Mesa 19.0 but version 19.1 is the minimum for full VC6 compatibility.

m4xw:
I used Mesa 19.1 rc1 on Ubuntu bionic aarch64, compiled from source and yes it showed.
that was with yesterday's branch.

fzurita:
@m4xw I have no idea add to the effects of calling eglInitialize multiple times. I fear it may reset all the configuration we perform here:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/mupen64plus-ui-console/ae-bridge/ae_vidext.cpp#L111

It may even be driver dependent.

hhromic:
After discussing with @m4xw , the solution to cover most cases is now in PR #94.
It uses `eglInitialize()` and also checks for `extensions == nullptr`. If issues appear, we can fix them later.

m4xw:
Any update?
I will close this otherwise... again.

swingflip:
will deal

m4xw:
@swingflip

m4xw:
@swingflip @evilwraith Last call here.

Issue #40, opened 2019-03-24
Long loading times before content starts for some Platforms


Ryunam:
I am currently testing this core out with the latest code from the GLideN64 branch and I'm noticing really long loading times when starting any content. 
I am not sure if such a delay is normal compared to most other cores which are usually instant, but with this one on my system I am having 1-2 minutes of waiting time after selecting 'Run' before the content actually starts.
This is happening with my usual RA config file as well as with a fresh new one.

sweatypickle:
Windows? No it's not normal as it doesn't happen with standalone mupen and GlideN64. However the delay I get I timed at 22-30 seconds above standalone.

m4xw:
Dont report issues for features until they are finished, unless I ask for it.

Ryunam:
Since you had distributed the core on Twitter, I thought this would be open and useful for reporting such things. Nevermind then.

m4xw:
We are not at that stage yet.
Also I am like the first guy to notice the issues, do you really think I wouldn't notice increased load time?

Ryunam:
It is not my intention to stir up an argument, so would you perhaps drop this unnecessarily aggressive act?

Second, as explained above I was under the impression that since you had distributed a build of this same core in a public fashion it would be perfectly reasonable to report and track ANY existing issue here: a lot of people do it and it can be beneficial for development. 
In fact, I can see that other threads have been opened here already to document either feature requests (the native resolution aspect, for example) or bugs, so I still do not really understand why you felt the need to close mine and get all authoritarian on this.

I can assure you it wasn't clear that reporting even something like an extended load time - a perfectly legitimate thing to report, even if you're actually the first guy to notice - was so frowned upon. Once again, nevermind.

m4xw:
@Ryunam I didn't want to be percieved as aggressive, in fact it's far from it (for my terms), so excuse me for that.
I am just really straight forward, also I didn't want to close the Issue without giving you a minimalistic explanation.
Issues are welcome, but the 2 specific issues you posted, I am well aware of and these are already in progress of being solved.
I mostly work on my internal tracker anyway.

m4xw:
Re-opening the Issue since it still happens on some platforms.
@mudlord didn't you say it didnt happen on newer gcc vers?
Weirdly enough I have this Issue not on Android aarch64 but on aarch32, yet toolchain ver should be the same (4.9).
Doesn't happen on Switch either.

mudlord:
[...]

PR #93, opened 2019-09-05
Fix for 20 second pause on some systems.

mudlord:
https://github.com/libretro/mupen64plus-libretro-nx/issues/40

Capture

This is a fix for the current 20 or so second loading pause on startup.

The culprit is the usage of srand() and rand(). Not using these functions and using a simple LSFR based number generator makes a massive performance increase and almost eliminates the pause. Its recommended that the function is inlined.


m4xw:
Thanks for doing this, long time outstanding issue!
So the issue was not enough entropy?
Will review it after the conflict is resolved!
Also don't mind these trolls, I personally don't give a f what they say ;)

mudlord:
Issue with doing on GPU was not enough entropy without using integer bitmath and SM4 features.

Using a simple LSFR on CPU is enough to give decent results.

GLideN64/src/NoiseTexture.cpp
```diff
@@ -34,27 +34,24 @@ NoiseTexture::NoiseTexture()
...
-u32 Rand(u32 rand_value)
+inline u32 irand()
 {
-#ifdef MINGW
-	rand_s(&rand_value);
-#else
-	rand_value = rand();
-#endif
-	return rand_value;
+        static u32 iseed = 1;
```
m4xw: Mind fixing the space?

mudlord: fixed?

m4xw: ye

m4xw:
@mudlord If you don't mind I will hold off merging this PR until I tried it on Switch this evening.
Want to squash it yourself or want me to do it on merge?
Also can u provide a comparison screenshot for reference? Will be like 7h until I can try myself.

mudlord:
do it on merge, and okay.

mudlord:
srand
lsfr

Results are almost identical.

m4xw:
Nice, thanks!

twinaphex:
Outstanding work. This really stuck out like a sore thumb on Windows.

mudlord:
No doubt, I was just extremely surprised why no one bothered to debug thoroughly **why** things were so slow. 

Didn't make sense to use a cryptographically secure RNG which uses thread local storage (on Windows at least, don't know how the CRT implements under Mac/Linux kernel), when a super simple LSFR based gen would suffice.

m4xw:
> why no one bothered to debug thoroughly why things were so slow.

Got other things to do + I only use Switch actively where this Issue doesn't exist, since we feed extra entropy there :P
If u look to blame someone, blame my laziness ;)

mudlord:
that and prob the Switch does things differently when it comes to RNG.

m4xw:
Well it provides hwrng, however its only used to seed newlib sufficiently iirc.

mudlord:
hwrng? nice.....is it the shit show that intel's hwrng was?

m4xw:
Its a decent API by the system.

m4xw:
Decided to merge it anyway already.

PR #106, opened 2019-10-06
Added core option: DisableExtraMem

RiskyJumps:
I added a core option for disable extra memory expansion.

### DIFF
--- a/GLideN64/src/Config.h
+++ b/GLideN64/src/Config.h
@@ -56,6 +56,7 @@ struct Config
 		u32 enableFragmentDepthWrite;
 		u32 enableBlitScreenWorkaround;
 		u32 hacks;
+		u32 disableExtraMem;
 #if defined(OS_ANDROID) || defined(OS_IOS)
 		u32 forcePolygonOffset;
 		f32 polygonOffsetFactor;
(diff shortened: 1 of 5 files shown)


m4xw:
This can't be added to gliden config stuff because it is mupen core specific, we also have a angrylion branch btw.
Anyway 5am soon here, will review in 10h.

custom/GLideN64/mupenplus/Config_mupenplus.cpp
```diff
@@ -119,7 +119,8 @@ extern "C" void Config_LoadConfig()
 	config.generalEmulation.enableLegacyBlending = enableLegacyBlending;
 	config.generalEmulation.enableNoise = EnableNoiseEmulation;
 	config.generalEmulation.enableLOD = EnableLODEmulation;
-	
+	config.generalEmulation.disableExtraMem = DisableExtraMem;
```
m4xw: Remove

mupen64plus-core/src/main/main.c
```diff
@@ -1003,7 +1003,8 @@ m64p_error main_run(void)
 
     count_per_op = CountPerOp;
     disable_extra_mem = ROM_PARAMS.disableextramem;
-
+    if (DisableExtraMem == 1)
+        disable_extra_mem = 1;
```
m4xw: do a \n before and after this block

GLideN64/src/Config.h
```diff
@@ -56,6 +56,7 @@ struct Config
 		u32 enableFragmentDepthWrite;
 		u32 enableBlitScreenWorkaround;
 		u32 hacks;
+		u32 disableExtraMem;
```
m4xw: Remove this

PR #110, opened 2019-10-08
Added core option: ForceDisableExtraMem

RiskyJumps:
I added a core option for disabling extra memory, according to given requirements.


m4xw:
Thanks

RiskyJumps:
Thanks to you for your time.

hhromic:
Hi guys, sorry for the late comment :)

What is the difference between setting this new option to `true` vs setting all/any existing `pX-pak` options to `none` or any non-`memory` value ?

hhromic:
Ah you are right, forgot there was a RAM expansion module != controller memory pack. My bad! :)

PR #111, opened 2019-10-09
Makefile fixes

MastaG:
- Add an option HAVE_LTCG we can set to enable LTO
- Optimize CPUFLAGS for Odroid-XU3/XU4
- Make sure we can override the BOARD variable (useful when cross-compiling or when running a mainline kernel which does not list the model in /proc/cpuinfo)
- Also add the CPUFLAGS when linking, this is required when the compiler/toolchain doesn't set the same default options
- Change default optimization from O2 to O3

This makes it possible to build with e.g.:
make HAVE_LTCG=1 ARCH=arm platform=odroid BOARD=ODROID-XU4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ FORCE_GLES=0
 FORCE_GLES3=0
 LLE=0
 
+HAVE_LTCG ?= 0
 DYNAFLAGS :=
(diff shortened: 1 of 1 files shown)


m4xw:
? Why the close

PR #114, opened 2019-10-09
Makefile fixes

MastaG:
Makefile improvements

- Add an option HAVE_LTCG we can set to enable LTO.
- Optimize CPUFLAGS for Odroid-XU3/XU4.
- Make sure we can override the BOARD variable (useful when cross-compiling or when running a mainline kernel which does not list the model in /proc/cpuinfo).
- Also add the CPUOPTS, PLATCFLAGS and CPUFLAGS when linking, this is required when the compiler/toolchain doesn't set the same default options as set when compiling.
- Change default optimization from O2 to O3.

This makes it possible to build with e.g.:
make HAVE_LTCG=1 ARCH=arm platform=odroid BOARD=ODROID-XU4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ FORCE_GLES=0
 FORCE_GLES3=0
 LLE=0
 
+HAVE_LTCG ?= 0
 DYNAFLAGS :=
 INCFLAGS  :=
 COREFLAGS :=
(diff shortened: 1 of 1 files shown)


m4xw:
You know u can just force push :P

MastaG:
yeah sorry about that.. I'm still a bit new to git.
Anyways, I've tested the changes on my XU3 and it doesn't break or become unstable.
However if you think the default optimization should be kept to O2, then I'll try to "force push" to this PR without the O3 commit.

m4xw:
Nah O3 is fine and its well tested on updated gcc.
Can you try your changes on the experimental branch too?

MastaG:
Sure, if you can give me a couple of hours (have to pick up the kids) I'll try it out and let you know.

MastaG:
The experimental branch with my PR on top of it works like a charm!
I've built with HAVE_LTCG=1 (-flto) enabled as well.
I don't know whether it's the cflags or you've put some more optimization into the experimental branch but it performs really reallty well.
I'm running RetroArch on top of gnome 3.34 wayland (Fedora 31) and I can still play super mario 64 in 720p without slowdowns.
As far as I can remember this wasn't possible using the mupen64plus build that came with batocera.linux the last time I tried it.

So I guess you can merge this PR and then cherry-pick it into experimental as well.

m4xw:
Thanks

PR #123, opened 2019-10-26
[WIP] IOS support

jet082:


m4xw:
Rebased + cleaned up on the IOS branch
https://github.com/libretro/mupen64plus-libretro-nx/commits/ios

m4xw:
I should get this on develop before Christmas, my schedule has been a bit... rough.

estefan3112:
Just to thank you for a very good iOS Core im RA 18.2 - hires textures work as they should, and with some filtering, it is an absolute eye candy on my iPad! Only save states for OOT cannot convert, so I am starting all over again - but it is so nice anyway!

m4xw:
"Things happened" and I didn't move my ass to clean it up yet.
Still have it on my watch, sorry for keeping it open for _too_ long.

m4xw:
Rebased the branch, @jet082 I'd still like to solve the aarch64 integration because rn it fails and the bot considers both the arm and aarch64 version failed.
Could you look into it on my ios branch?

estefan3112:
Hi m4xw, I just completed OOT Master Quest on my iPad with this core, and it was absolutely great. Do you accept reports on the iOS branch? There are some glitches on static screens after some time, like this, but nothing really to concern. Gameplay is perfect! So again thanks!

Legend of Zelda, The - Ocarina of Time - Master Quest (U)  ! -200126-113222

m4xw:
Does this happen only with high res texture?
Great that interpreter was enough :p

estefan3112:
> Does this happen only with high res texture?

Yes, it happens only if hires is enabled. The behaviour is gone if I switch to lores and restart. The workaround in hires mode is to delete the memory cache file of the game, then it runs normal again, and sometimes the problem starts again. So basically nothing to be worried about.

m4xw:
@estefan3112 What if you generate the texture cache on the device?
I assume u used a pre-built.

estefan3112:
@m4xw - well does this work? I thought that only project64 allows building an own htc - also on my iPad with mupen64plus-nx? Would love to do this!

m4xw:
@estefan3112 Yes, same hires_ folder as pj64 from Mupen64Plus system/ folder.
It will be stuck on black screen until its done - no notification.

estefan3112:
I will try it this weekend! Thanks!

m4xw:
Rebased earlier with the rsp updates

estefan3112:
> @estefan3112 Yes, same hires_ folder as pj64 from Mupen64Plus system/ folder.
> It will be stuck on black screen until its done - no notification.

Oh yes, this works perfectly. This is nowhere documented, can I help documenting this function, it is rather indispensable!

m4xw:
> 
> 
> > @estefan3112 Yes, same hires_ folder as pj64 from Mupen64Plus system/ folder.
> > It will be stuck on black screen until its done - no notification.
> 
> Oh yes, this works perfectly. This is nowhere documented, can I help documenting this function, it is rather indispensable!

In fact its wrongly documented since forever in the old mupen core, I don't have a specific mupen next document yet in the libretro docs iirc.
Theres however some different behaviour with windows apparently but thats likely some folder path bullshit.
However i'd like to add some notifications at some point..
https://github.com/libretro/docs
This was the docs page of the old mupen core, it would need a remake https://github.com/libretro/docs/blob/master/docs/library/mupen64plus.md

estefan3112:
> In fact its wrongly documented since forever in the old mupen core, I don't have a specific mupen next document yet in the libretro docs iirc.
> Theres however some different behaviour with windows apparently but thats likely some folder path bullshit.
> However i'd like to add some notifications at some point..
> https://github.com/libretro/docs
> This was the docs page of the old mupen core, it would need a remake https://github.com/libretro/docs/blob/master/docs/library/mupen64plus.md

Some documentation is now already in, I kept it very short, as a start.

The .htc creation also works on MacOS with standalone core and self-compiled video plugin.
But on both Mac and iPad, one of the Djipi packs apparently misses the correct fonts. More or less all fonts are scrambled. But that's limited to certain hd packs, others work fine.

Screenshot 2020-02-08 at 12 06 44

m4xw:
Close in favor of #149

PR #132, opened 2019-12-15
Emscripten support

ToadKing:
# NOTE: PLEASE GIVE IT A GOOD TEST BEFORE MERGING!
This changed a lot of core mupen code to yank out libco support and may have broken things, especially with restarts and loading multiple games and such.

image
_It's still a work in progress_

Other than the libco changes, it also has a view minor fixes for compiling. Also emscripten compiles to an archive file instead of bytecode file, due to issues with wanting memory set when linking. The RetroArch build will handle that instead.


ToadKing:
closing for now, conflicting with some upcoming changes

PR #150, opened 2020-02-19
Add support for loading GL symbols using dlsym() instead of libretro API

hhromic:
* Required for platforms with EGL version < 1.5
* To enable, use new compile-time define: `-DGL_USE_DLSYM`

Currently only enabled for Raspberry Pi platforms that use the legacy Broadcom driver.

### DIFF
--- a/GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
+++ b/GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
@@ -5,11 +5,16 @@
 
 #include "GLFunctions.h"
 
-// For Libretro
 #define ASSIGN_PROC_ADR(proc_type, proc_name) ptr##proc_name = gl##proc_name
+#if defined(GL_USE_DLSYM)
+// Use dlsym() to load GL symbols from the default shared object search order
+#define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) dlsym(RTLD_DEFAULT, "gl"#proc_name)
+#else
+// Use libretro API to load GL/EGL symbols
 #define glGetProcAddress glsm_get_proc_address
 #define GL_GET_PROC_ADR(proc_type, proc_name) ptr##proc_name = (proc_type) glGetProcAddress("gl"#proc_name)
 #define GL_GET_PROC_ADR_EGL(proc_type, proc_name) ptr##proc_name = (proc_type) glGetProcAddress("egl"#proc_name)
+#endif
 
 #if 0
 #ifdef OS_WINDOWS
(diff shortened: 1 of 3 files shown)


GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
```diff
@@ -398,4 +410,8 @@ extern "C" void initGLFunctions()
 
 	GL_GET_PROC_ADR(PFNGLEGLIMAGETARGETTEXTURE2DOESPROC, EGLImageTargetTexture2DOES);
 	GL_GET_PROC_ADR(PFNGLEGLIMAGETARGETRENDERBUFFERSTORAGEOESPROC, EGLImageTargetRenderbufferStorageOES);
+
+#if defined(USE_DLSYM_FOR_GL)
+	dlclose(dlhandle);
```
m4xw: Does this actually work?
Probably only because frontend still has a handle, theoretically this would unload the GLES lib if all other handles are closed.

hhromic: Yes yes, I tested it of course.
Nothing theoretic, that is exactly how it works. `dlclose()` just decreases usage count number.

hhromic:
Btw, is there a function to cleanly exit on error from gliden? Right now if the lib is not found it logs the error but the plugin will still crash afterwards. I assume `exit()` is no bueno here?

hhromic:
New version of the PR now. I found out that `dlsym()` can be called with `RTLD_DEFAULT` as the handle parameter and will search symbols in all the linked libraries, including the GL lib. We don't need to supply a hard-coded path nor use `dlopen()`/`dlclose()` anymore thus greatly simplifying the fix.

> RTLD_DEFAULT
> Find the first occurrence of the desired symbol using the default shared object search order. The search will include global symbols in the executable and its dependencies, as well as symbols in shared objects that were dynamically loaded with the RTLD_GLOBAL flag.

m4xw:
@hhromic Can you please target the 2.0.5-RC1 branch, then I can merge it (will need some changes for thr branch anyway)
Also please amend and remove the @ next to my name.

hhromic:
@m4xw rebased and commit msg amended. Test pending.
I'm going to test this change and the `v2.0.5-RC1` later tonight on my RPI3B with legacy drivers to confirm all is fine to merge. Thanks!

Ps: I never liked the `YOLO_DLSYM` name for the define, I think this core deserves seriousness in its variable names :) Therefore I renamed it to `GL_USE_DLSYM`. I hope you don't mind.

m4xw:
Whatever works, just give me the go when you are ready.

hhromic:
Finished testing it now. All is working on RPI3B + Legacy Drivers :+1: 

I tested the rebased version (`v2.0.5-RC1` branch) and the core is working pretty much as usual for this setup.

m4xw:
Ty

PR #160, opened 2020-03-14
[Reference] GLES2 doesn't support Async color copy

dankcushions:
HAVE_OPENGLES is enabled on GLES3 builds: https://github.com/jdonald/mupen64plus-libretro-nx/blob/develop/Makefile.common#L377

currently this incorrectly defaults this option off for GLES3, when it's only GLES2 devices that can't use it in async mode: https://github.com/gonetz/GLideN64/blob/0a5216f8e1e8dad7e73b9eaf32082b2313d2c501/src/Graphics/OpenGLContext/opengl_GLInfo.cpp#L141

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -272,7 +272,7 @@ static void setup_variables(void)
         { CORE_NAME "-EnableLODEmulation",
             "(GLN64) LOD Emulation; True|False" },
         { CORE_NAME "-EnableCopyColorToRDRAM",
-#ifndef HAVE_OPENGLES
+#ifndef HAVE_OPENGLES2
             "(GLN64) Color buffer to RDRAM; Async|Sync|Off" },
 #else
             "(GLN64) Color buffer to RDRAM; Off|Async|Sync" },


twinaphex:
Hi @m4xw, let me know if this can be merged.

m4xw:
This was done on purpose.
I will reevaluate.

m4xw:
@dankcushions What is your test case/device to think that the average GLES3 device can handle it by default?
I guess we should change the GLES2 case anyway and remove it completely and default to off for GLES3.
I am aware this breaks stuff, but where its strictly required the GLideN64 ini should take over and force the settings, the GLES2 limitations are still in place inside GLideN64.
So.. dunno, thoughts?

dankcushions:
> @dankcushions What is your test case/device to think that the average GLES3 device can handle it by default?

pi4 is 3.1 and uses mesa (which i'm assuming is pretty vanilla). i don't know how representative that is, but again i point to the original GLideN64 code which only disables it for GLES2: https://github.com/gonetz/GLideN64/blob/0a5216f8e1e8dad7e73b9eaf32082b2313d2c501/src/Graphics/OpenGLContext/opengl_GLInfo.cpp#L141, and it's default for everything else - if they think it's safe I think we can!

> I guess we should change the GLES2 case anyway and remove it completely and default to off for GLES3.

it's async by default for GLideN64 standalone: https://github.com/gonetz/GLideN64/blob/e7e6061011615a3aa6cea0d02ac1d4f08a39d3fe/ini/GLideN64.ini#L27, so i think we should default to on (btw it's disabled for some games in the custom.ini https://github.com/gonetz/GLideN64/blob/dff219f26f8fd3c4650743c10899b1912ee43ccf/ini/GLideN64.custom.ini#L15).

> I am aware this breaks stuff, but where its strictly required the GLideN64 ini should take over and force the settings

do you mean that GLideN64.custom.ini? because that does not contain any entries for CopyToRDRAM=2 (async) because it's the default. it used to, but they were all removed once it became the default.

dankcushions:
this may actually be where the defaults are kept: https://github.com/gonetz/GLideN64/blob/master/src/Config.cpp#L66 (either way, same thing - defaults to 2 - async)

i'm not sure whether GLideN64.ini or this dictates the defaults for standalone, but yeah.

m4xw:
Guess I will re-work the option, we currently dont expose triple buffering either, might be a good time to catch that all.

PR #173, opened 2020-04-10
Fix F-Zero X Expansion Kit cartridge hack by using F-Zero X microcode

dmrlawson:
Made against this project as requested.

As far as I know it needs to be treated the same as regular F-Zero X, so this change ensures that.

### DIFF
--- a/mupen64plus-rsp-hle/src/hle.c
+++ b/mupen64plus-rsp-hle/src/hle.c
@@ -228,7 +228,7 @@ static bool try_fast_audio_dispatching(struct hle_t* hle)
             case 0x1f701238: /* Mario Artist Talent Studio */
                 alist_process_nead_mats(hle); return true;
             case 0x1f4c1230: /* FZeroX Expansion */
-                alist_process_nead_efz(hle); return true;
+                alist_process_nead_fz(hle); return true;
             default:
                 HleWarnMessage(hle->user_defined, "ABI2 identification regression: v=%08x", v);
             }


dmrlawson:
How come?

m4xw:
@dmrlawson It got autoclosed because I merged the branch, didn't close it on purpose.
Like I said tho, it should still use the efz func and we should return -1 in the fallback codepath.

m4xw:
.. Also I should point out I don't want to make the change as is, because I will totally forget about it and then wonder why LLE fallback isn't working once implemented, lol.
You very first suggestion in Discord was acceptable, now that I know the usage of the func.
Hope you don't mind.

dmrlawson:
Ah, fair enough.

My understanding is that returning `-1` from the `HleForwardTask` thing means that it will use LLE in a number of different places, killing performance. There's no need to use it for F-Zero X Expansion because afaik it's the same ucode as F-Zero X.

Have I misunderstood something somewhere?

dmrlawson:
Or maybe I'm totally misunderstanding and the reason that function gets hit is that the microcode _is_ different, but sufficiently similar that it mostly works

m4xw:
-1 should mean LLE Fallback is not implemented
0 should mean forwarded to LLE RSP.
At least from the quotes that I saw.

m4xw:
(Basically just restore some code I nuked long ago)

dmrlawson:
If that's the case, how is it working currently when LLE is not implemented? That function is checked in a few different places (although maybe only when other options have been exhausted?)

m4xw:
As you can see its not working.
It thinks RSP command was forwarded while it only got ignored.

dmrlawson:
Sorry - I mean with other, standard games

m4xw:
Almost all ucodes are implemented in the HLE RSP.
So it almost never happens to begin with.

PR #182, opened 2020-04-30
Always return -1 from HleForwardTask

dmrlawson:
We don't have LLE yet so this means we don't try to forward things to an LLE implementation. Fixes F-Zero X Expansion


m4xw:
ty

PR #185, opened 2020-05-04
Fix rebase build

dmrlawson:


m4xw:
👍 will go from here as discussed in discord

PR #187, opened 2020-05-04
Update libretro_core_options.h

Tatsuya79:
add default for v2 option, add some explanations, moving values around.

### DIFF
--- a/libretro/libretro_core_options.h
+++ b/libretro/libretro_core_options.h
@@ -593,7 +593,7 @@ struct retro_core_option_definition option_defs_us[] = {
     {
         CORE_NAME "-MaxTxCacheSize",
         "Max texture cache size",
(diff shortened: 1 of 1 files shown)


hhromic:
This is great! 🎉 

Shouldn't the default options be based on #ifdefs as previously done in the old options?

For example in

https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L301-L307

it can be seen that for certain platforms the default texture cache size limit is different.

m4xw:
@hhromic its all still WIP, ifdef's and platform defaults are still missing so is hiding options + some more

hhromic:
Ah that's great then. Just thought it might have been forgotten :)
Thanks for all the good work! Cheers!

PR #190, opened 2020-05-06
Update xxhash to the version in core

dmrlawson:
As discussed, use a subrepo for xxhash so that we use the same version for GLideN64 and mupen64plus-core (and anything else).

It seems a bit overkill to pull in the whole repo but I'm not sure if there's a cleaner solution.


m4xw:
Yea lets not do that, I can just keep it manually updated.
This pulls in too many noise, lets have it use the variant in the root folder tho for each project.

dmrlawson:
Fixed

PR #191, opened 2020-05-06
Remove CRC.cpp and use GLideN64 implementations

dmrlawson:
**I think this probably needs work**

I'm going by the logic in https://github.com/libretro/GLideN64/blob/develop/src/CMakeLists.txt#L336 and assuming that we want the neon implementation where we have neon available. This was only a quick attempt and I've only tried building on x86_64.

Do we have an architecture variable available in the makefile?


dmrlawson:
Fails to build for rpi, having another look.

m4xw:
armv8 crc instructions should be faster than a neon impl

PR #192, opened 2020-05-07
Core options ifdefs

dmrlawson:
I think I've got all of the for the main options. Someone should double check the logic is correct.

There seems to be a half followed convention of commenting what the `#endif` is about, shall I add that or not?

### DIFF
--- a/libretro/libretro_core_options.h
+++ b/libretro/libretro_core_options.h
@@ -36,28 +36,6 @@ extern "C"
 #define CORE_NAME "mupen64plus"
 #endif
 
-// Option entries
-#define OPTION_ENTRY_RDP_GLIDEN64 "gliden64"
-#define OPTION_ENTRY_RSP_HLE "hle"
-
-#ifdef HAVE_THR_AL
-#define OPTION_ENTRY_RDP_ANGRYLION "|angrylion"
-#else
-#define OPTION_ENTRY_RDP_ANGRYLION ""
-#endif // HAVE_THR_AL
-
-#ifdef HAVE_PARALLEL_RSP
-#define OPTION_ENTRY_RSP_PARALLEL "|parallel"
-#else
-#define OPTION_ENTRY_RSP_PARALLEL ""
-#endif // HAVE_PARALLEL_RSP
-
-#ifdef HAVE_LLE
-#define OPTION_ENTRY_RSP_CXD4 "|cxd4"
-#else
-#define OPTION_ENTRY_RSP_CXD4 ""
-#endif // HAVE_LLE
-
 struct retro_core_option_definition option_defs_us[] = {
     {
         CORE_NAME "-cpucore",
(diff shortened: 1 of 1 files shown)


libretro/libretro_core_options.h
```diff
@@ -224,7 +237,11 @@ struct retro_core_option_definition option_defs_us[] = {
             {"Sync", NULL},
             {NULL, NULL},
         },
+#ifndef HAVE_OPENGLES
         "Async"
```
dankcushions, line 241: I think this logic is slightly wrong - see https://github.com/libretro/mupen64plus-libretro-nx/pull/160/files

libretro/libretro_core_options.h
```diff
@@ -304,8 +325,13 @@ struct retro_core_option_definition option_defs_us[] = {
             {"True", NULL},
             {NULL, NULL},
         },
+#ifdef HAVE_OPENGLES
+        "False
+#else
         "True"
```
dankcushions, line 331: this is enabled by default on upstream - https://github.com/gonetz/GLideN64/blob/a96853d559680cca17fdc4dfa5891dad03d79b66/src/Config.cpp#L58 - safe? works on my GLES2 and 3 devices

libretro/libretro_core_options.h
dmrlawson: This is just copied from here: https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L274

I would suggest that this MR simply replicates the current state and the other merge request deals with the behaviour change, but I'm open to doing it here if that's what's desired.

libretro/libretro_core_options.h
dmrlawson: This is from https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L293-L294

Again I'd suggest a separate MR for changing behaviour.

dmrlawson:
Also, the `EnableNoiseEmulation` option needs to be removed as it's not configurable in the latest GLideN64, but I left it for now as this is to merge into the `improved_coreoptions` branch which doesn't have it yet

dankcushions:
oh sorry, didn't realise this wasn't for main branch 👶 
we can revisit this stuff later

m4xw:
We will Incooperate your feedback, dw.
didn't forget about it

m4xw:
CXD4 and Parallel are opt in https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L244-L246
https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L95-L116
Using some macro trickery.
https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L265-L270
https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/libretro/libretro.c#L345-L354

dmrlawson:
Fixed. Got rid of the macros because I don't think they're necessary now

m4xw:
Gonna merge for now and do manual fixup if required.

PR #195, opened 2020-05-11
Staging PR for Mupen64Plus-Next v.2.1.0

m4xw:
This PR might be manually merged and is just intended as overview.

PR #200, opened 2020-05-20
Fix GCC10 build

heuripedes:
GCC behavior became more strict in version 10.
>**Default to -fno-common**
>A common mistake in C is omitting extern when declaring a global variable in a header file. If the header is included by several files it results in multiple definitions of the same variable. In previous GCC versions this error is ignored. GCC 10 defaults to -fno-common, which means a linker error will now be reported. To fix this, use extern in header files when declaring global variables, and ensure each global is defined in exactly one C file. If tentative definitions of particular variables need to be placed in a common block, __attribute__((__common__)) can be used to force that behavior even in code compiled without -fcommon. As a workaround, legacy C code where all tentative definitions should be placed into a common block can be compiled with -fcommon.

      int x;  // tentative definition - avoid in header files
      extern int y;  // correct declaration in a header file
  

https://gcc.gnu.org/gcc-10/porting_to.html

Thanks to RealNC for pointing it out.

### DIFF
--- a/mupen64plus-core/src/device/r4300/idec.h
+++ b/mupen64plus-core/src/device/r4300/idec.h
@@ -82,6 +82,6 @@ size_t idec_u53(uint32_t iw, uint8_t u53, uint8_t* u5);
 
 #define IDEC_U53(r4300, iw, u53, u5) (void*)(((char*)(r4300)) + idec_u53((iw), (u53), (u5)))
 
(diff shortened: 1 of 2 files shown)


m4xw:
I can't accept this, it's already fixed on the rebase branch and until that is merged no further commits will happen on develop.
Thanks tho!

m4xw:
https://github.com/libretro/mupen64plus-libretro-nx/pull/195

PR #205, opened 2020-05-29
Fix double free and multiple definition errors

deltabeard:
This fixes a segmentation fault which occurs when the frontend does not support config environment commands.
This fixes a multiple definition issue which produces an error during compile with GCC 10.1.
Tested working.


deltabeard:
It was requested that a failed environment callback produce an error and have the core exit rather than continuing with a default setting.
A fix for multiple definition compile errors is already pending.

PR #211, opened 2020-06-15
Add enable native res factor core option

dmrlawson:


dankcushions:
i don't think this is working quiet correctly as it still seems to care about the 4:3/16:9 resolution core options. for example:
```
mupen64plus-next-43screensize = "320x240"
mupen64plus-next-EnableNativeResFactor = "1"
mupen64plus-next-HybridFilter = "False"
```
(turn off hybrid filter because it obfuscates the scaling issues)
=
007 - GoldenEye (USA)-200616-094029

i don't know how the native resolution factor is being applied, but it still seems to be outputting 320x240 as the final resolution to retroarch (which is then scaled up to display resolution), negating the benefit.

how i would expect this option to work is for it to at least ignore the screensize options, and the final resolution sent to retroarch to be the native-resolution of the game in question.

i would personally prefer to remove the screensize options entirely as a scaling of native res makes more sense in a retroarch context.

i would also get rid of the hybrid filter option as that is only required for smoothing out the scaling from render to display resolution, and retroarch itself controls that operation, so it should be implemented as a retroarch shader. however, not strictly part of this issue so feel free to ignore :P

dmrlawson:
My understanding is that though the internal frame buffer is 320x240, there is some other step that projects that onto something close to 640x480. So running it at 320x240 would basically be downscaling from the native resolution. Indeed, if you scale that example image you have down to 320x240 you get this:
image

I may have some misunderstanding here though so I'll let someone else comment.

dankcushions:
so for the goldeneye example, the native resolution of that goldeneye intro/menu screen is, i'm told, 440x330. what this core seems to be doing with the options as i describe above is

1. render 440x330 (`EnableNativeResFactor = "1"`)
2. scale that to 320x240 (`43screensize = "320x240"`)
3. scale to display resolution (in my case, 1080p) via retroarch. 

what i'm proposing is that we should get rid of step 2. when EnabledNativeResFactor is enabled (or perhaps, just get rid of it entirely). this is what GLideN64 does with the equivalent options - it scales the native resolution image to your display resolution.

dmrlawson:
Right, I think see what you mean.

This change is just focused on allowing users to control the resolution factor passed to GLideN64. We already pass the option, but the value is hardcoded.

I think you can achieve what you want by setting the Native Resolution Factor to 1 and your resolution to 1920x1080, rather than 320x240, right? Then it renders at native res and scale directly to 1920x1080

dankcushions:
> I think you can achieve what you want by setting the Native Resolution Factor to 1 and your resolution to 1920x1080, rather than 320x240, right? Then it renders at native res and scales directly to 1920x1080

not quite - 1920x1080 is not an available resolution, but 1440x1080 is. i tried that option and it still produces a better, but still bad image (look at the top row of text in particular):
007 - GoldenEye (USA)-200616-113952

which makes me think the option isn't quite working properly?

i should say that even if this work-around worked, it wouldn't be practical for a generic default setup for something like retropie (my concern) as we wouldn't know the users display resolution ahead of time. but i can agree that fixing it beyond could go beyond the scope of the PR.

m4xw:
Scheduled for manual merge, nativeRes viewport stuff will go on the nativeResBranch for now, the core option will be added to upcoming v2.1

m4xw:
@dankcushions give it a shot

dmrlawson:
Removed the native res branch stuff as it's not scheduled for the next release and seems to do weird things in my testing.

I've pushed it to a different branch here: https://github.com/dmrlawson/mupen64plus-libretro-nx/tree/rebase_native_res

m4xw:
Gonna merge as-is for now

PR #213, opened 2020-06-20
Fix order of steps to close a rom and shutdown the core

88hcsif:
Libretro frontends will unload the rom first and then deinit the core.
This is generally not a big deal since most (all?) libretro frontends will completely unload the libretro library loaded dynamically before loading it again and starting a new game.
There are actually other issues that prevent a libretro frontend from just starting a new game without unloading the shared library, multiple static variables (I only found a few) will not reset and the core will be in a bad state and crash.
However, I think this is still beneficial to make sure there are possibly fewer memory leaks.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -417,10 +417,11 @@ static bool emu_step_load_data()
 
     log_cb(RETRO_LOG_DEBUG, CORE_NAME ": [EmuThread] M64CMD_ROM_OPEN\n");
 
-    if(CoreDoCommand(M64CMD_ROM_OPEN, game_size, (void*)game_data))
+    ret = CoreDoCommand(M64CMD_ROM_OPEN, game_size, (void*)game_data);
+    if (ret)
     {
         if (log_cb)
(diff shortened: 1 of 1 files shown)


m4xw:
This needs some consideration for static platforms and already breaks on the upcoming relase on the `rebase` branch.

88hcsif:
Sounds good, I will change the patch to be based on the rebase branch.
And I will ad a few more details as to why I think this makes sense.
But if there are platform that link the library statically and rely on this order it would definitely need some testing...
Also the rebase branch is triggering a segmentation fault when closing the game in the frontend I am using when it tries to call retro_serialize, so I will investigate that too...

m4xw:
>Also the rebase branch is triggering a segmentation fault when closing the game in the frontend I am using (https://github.com/giongto35/cloud-game) when it tries to call retro_serialize, so I will investigate that too...

Probably because the threaded renderer changes, guess it doesnt like to advance the core in serialize with that frontend, no way around it tho if you want savestates with thr renderer.

88hcsif:
Ah I see, I should be able to fix that in the frontend, I just need to make sure serialize is called from the thread that holds the openGL context.

Issue #146, opened 2020-02-12
MacOS Compilation Hurdles


estefan3112:
Before I start with Pull Requests, here are the first issues when trying to compile the master branch under MacOS Mojave. In the sequence of the three errors - two of them I think I could resolve by help of Google.

./GLideN64/src/PaletteTexture.h:3:10: fatal error: **'malloc.h' file not found**
**suggested change for MacOS:** #include <stdlib.h>

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp:28:10: fatal error: **'X11/Xutil.h' file not found**
**suggested addition for MacOS:** CPPFLAGS=-I/opt/X11/include
(I was unsure where to put that, the general makefile? If I executed it in the Terminal before compile, it was just fine9

Finally, that one I could not resolve by myself:
GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp:326:2: error: assigning to 'PFNGLTEXIMAGE2DMULTISAMPLEPROC' (aka 'void (*)(unsigned int, int, int, int, int, unsigned char)') from incompatible type
      'void (GLenum, GLsizei, GLenum, GLsizei, GLsizei, GLboolean)' (aka 'void (unsigned int, int, unsigned int, int, int, unsigned char)'): type mismatch at 3rd parameter
      ('GLint' (aka 'int') vs 'GLenum' (aka 'unsigned int'))
        ASSIGN_PROC_ADR(PFNGLTEXIMAGE2DMULTISAMPLEPROC, TexImage2DMultisample);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp:8:64: note: expanded from macro 'ASSIGN_PROC_ADR'
#define ASSIGN_PROC_ADR(proc_type, proc_name) ptr##proc_name = gl##proc_name
                                                               ^~~~~~~~~~~~~
<scratch space>:33:1: note: expanded from here
glTexImage2DMultisample
^~~~~~~~~~~~~~~~~~~~~~~
./libretro-common/include/glsm/glsmsym.h:165:37: note: expanded from macro 'glTexImage2DMultisample'
#define glTexImage2DMultisample     rglTexImage2DMultisample
                                    ^~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #216, opened 2020-06-21
Fix Mac compilation issues

88hcsif:
Fix the compilation hurdles mentioned in #146 and more.

I guess the change to GLideN64/src/PaletteTexture.h should be pushed upstream first.
However, according to the CI build status, https://github.com/libretro/GLideN64 already compiles on OSX...
I wonder if the travis environment include malloc.h some other way...


GLideN64/src/PaletteTexture.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 #include <memory>
 
-#ifdef IOS
+#if defined IOS || defined OSX
```
m4xw, line 4: `#if defined(IOS) || defined(OSX)`

libretro-common/glsm/glsm.c
```diff
@@ -2500,7 +2500,11 @@ void rglProgramBinary(GLuint program,
 
 void rglTexImage2DMultisample( 	GLenum target,
   	GLsizei samples,
+#if defined(OSX)
+  	GLint internalformat,
```
m4xw, line 2504: I need to double check them, I doubt it differs for IOS so its probably wrong type to begin with

estefan3112:
Hi, the core builds fine in MacOS Mojave, Retroarch loads a version 2.1 2dbff90, but each game crashes immediately. Crash report here?

88hcsif:
It works fine with the libretro frontend which is part of [cloud-game](https://github.com/giongto35/cloud-game).
Maybe retroarch also does not expect the retro_serialize function to run the run_retro?
But that would also imply that retroarch calls the retro_serialize function without being prompted (which might be true if that is how the rewind functionality works...).
@m4xw is retroarch the frontend you normally use?

m4xw:
I changed it to only do the co_switch earlier today, omitting the video cb

m4xw:
But also if that was the cause, then there is a bug somewhere else in the frontend.

88hcsif:
Okay, Retroarch has issues even with the code before the rebase, so it should not be related to the retro_serialize changes.
I think it is some issue with how OpenGL is initialised...

m4xw:
oh I guess i killed that PR on rebase by accident.
Can you rebase / reopen this and target develop?

PR #217, opened 2020-06-21
Close ROM after emulation is stopped

88hcsif:
Because of how mupen64pluc-core is structured, it is not possible to close a ROM until the emulation is stopped (M64CMD_ROM_CLOSE will simply be ignored).
Moving the logic to stop emulation from retro_deinit to retro_unload_game (it was already there for threaded renderer) and move the M64CMD_ROM_CLOSE command at the end.
Also add calls to CoreShutdown to free more memory. 

I think the fact that ROMs were not closing properly has not been an issue so far because most (all?) libretro frontends will completely unload any libretro library (loaded dynamically) before loading it again and starting a new game.
This fix does not suddenly allow frontends to start a new game without unloading mupen64plus first.
This is because there are other issues with static variables (I only found a few) that are not reset properly during deinitialization and that will leave mupen64plus in a bad state and crash.
However, I think this fix is still beneficial to make sure there are possibly fewer memory leaks.


m4xw:
I still plan on cherry picking this

PR #221, opened 2020-06-26
WIP: Native res stuff

dmrlawson:
Making a pull request to track progress of this. Not intended to be merged at the moment.

This branch is intended to add the stuff @m4xw wrote in the `native_res` branch but rebased on top of `develop`.

I copied in the @m4xw's stuff because the real rebase looked frightening.

It currently seems to work so long as `Overscan = Enabled/True`. Otherwise the image is stretched and off centre.

I'm looking into why this is at the moment.


dmrlawson:
As I understand, the overscan related issue is being discussed here: https://github.com/gonetz/GLideN64/issues/2301

If I understand correctly, it sounds like what I was coming to suspect, that with Overscan (i.e. cropping/removal of overscan) the rendered image is projected onto an additional frame buffer (which emulates the VI). This PR uses the resolution of the VI as the window resolution passed to Retroarch. With Overscan disabled this additional frame buffer isn't used so we get the wrong resolution passed to retroarch.

PR #226, opened 2020-06-28
Small changes to core option descriptions and titles.

bslenul:
* Changed "(GLN64)" and "(AL)" to "(GLideN64)" and "(Angrylion)", so it matches choices from the "RDP Plugin" option. Less confusing for new users IMO.
* Moved them from descriptions to title for clarity sake (especially if you have sublabels OFF), and for consistency with the new "(ParaLLEl-RDP)" options.
* Changed "disable" to "Disabled" for downsampling option.

Hope I did everything correctly :x

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -1276,7 +1276,7 @@ static void update_variables(bool startup)
         var.value = NULL;
         if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
         {
-            if (!strcmp(var.value, "disable"))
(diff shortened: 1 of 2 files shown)


m4xw:
All changes that you mentioned have been done explicitly on purpose.
I know this has some issues with the qt ui but thats a different story, changing it like you did was basically how I had it at first and figured it makes all even worse.
So I can't accept this PR.

bslenul:
OK, I think it makes core options nicer to navigate but you probably have some good reasons so I won't argue :D 

Thanks for looking at it anyway, and keep up the great job on that core ❤️

Issue #239, opened 2020-07-24
Missing geometry (or textures?) on GLES2 since commit 6469626


shantigilbert:
After commit 6469626a8fc8191518ae35543b80fd0002fd1604 I am getting a lot of missing textures and/or geometry, I tried random commits after that one and some of them just show white boxes (no texture or geometry at all) and the latest commit has a lot of glitched geometry and/or textures.

see the attached video: https://youtu.be/zU3Xpgi6H7w

also I get this on the log

```
[libretro INFO] mupen64plus: Initializing 4 RDRAM modules for a total of 8 MB
[libretro INFO] mupen64plus: Starting R4300 emulator: Dynamic Recompiler
[libretro INFO] mupen64plus: Init new dynarec
[libretro INFO] mupen64plus: ARM CPU Features:
[libretro INFO] mupen64plus: Enabled TLB
WARNING! Not implemented.
WARNING! Not implemented.
WARNING! Not implemented.
```
While on the last working commit (78233e1a925b8b0677918b0fe0894d56383e3923) I did not get the "WARNING!" 

```
[libretro INFO] mupen64plus: Initializing 4 RDRAM modules for a total of 8 MB
[libretro INFO] mupen64plus: Starting R4300 emulator: Dynamic Recompiler
[libretro INFO] mupen64plus: Init new dynarec
[libretro INFO] mupen64plus: Enabled TLB
```

Thank you!

m4xw:
What if you switch lod emu to any other value?

shantigilbert:
Same issue if I change LOD to false (was True). 

Also this seems to only happens with the 007 game? 
I tried Ocarina of time - seems working fine
Mario Kart - working fine
Mario 64 - working fine
I am sure there are other games affected, but I do not know which ones to test.

Although I still get the "WARNING! Not implemented." on the log

m4xw:
Are u using a GLES3 platform with a GLES2 build?

shantigilbert:
I am using `platform=AMLG12B` for arm and `platform=odroid64 BOARD=N2` for aarch64 to compile and I can see `   GL_LIB := -lGLESv3` being set, these have worked well before, but only after that commit that it broke. 

should I do the opposite and force -lGLESv2 ?

shantigilbert:
Ok so forcing both 
```
GLES3 = 1
-lGLESv3
```
On the Makefile fixes this issue. 

Thanks

m4xw:
[...]

PR #246, opened 2020-08-05
Use GLES3 for AMLG12/Odroid N2  platform

shantigilbert:
As discussed on https://github.com/libretro/mupen64plus-libretro-nx/issues/239

I tried to leave all the other platforms intact as I have no real idea how they are affected.


m4xw:
No Objections from fellow platform users?

PR #252, opened 2020-08-27
Add asm defines for Android x86 / x86_64 and allow Android x86_64 core to be built

farmerbb:
This allows the x86_64 core to be built for Android.  Tested on my Pixelbook and it works fine, though with lower performance than the ParaLLEl core.

This also adds the asm defines so that the x86 core could be built, but due to the x86 dynarec using text relocations, Android 6.0+ will refuse to run it, so I left x86 out of the makefile.


Jj0YzL5nvJ:
Does this change in upstream can help you? mupen64plus/mupen64plus-core#775

m4xw:
The issue with Android was I just couldn't get the Make target to trigger right, not the gawk dep itself.
Thanks @farmerbb will get to it the next time i do maintenance.

PR #256, opened 2020-09-03
Fix OSX build

kivutar:


m4xw:
Thats all?

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
```diff
@@ -342,7 +342,9 @@ extern "C" void initGLFunctions()
 	ASSIGN_PROC_ADR(PFNGLDELETEFRAMEBUFFERSPROC, DeleteFramebuffers);
 	ASSIGN_PROC_ADR(PFNGLGENFRAMEBUFFERSPROC, GenFramebuffers);
 	ASSIGN_PROC_ADR(PFNGLFRAMEBUFFERTEXTURE2DPROC, FramebufferTexture2D);
+#ifndef __APPLE__
 	ASSIGN_PROC_ADR(PFNGLTEXIMAGE2DMULTISAMPLEPROC, TexImage2DMultisample);
```
m4xw, line 346: Use ifdef apple GL_GET_PROC_ADR
If it doesnt resolve, let it null it via getproc
We don't wanna skip assignment here.

m4xw (changes requested):
See review

m4xw:
Also I expect u probably need to rewrite most of the mac osx makefile target.

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
m4xw: Actually nvm, assign proc adr should work since it assigns to `rgl` and there it's jst a nop.
Why was this changed?

m4xw:
bump

aschain:
Hi all, I appreciate the work on this for os x! I made the above changes to the three files mentioned, but was not able to get it to compile. I then also changed line 338 of the root Makefile from:
`   ASFLAGS = -f elf -d ELF_TYPE`
to:
`   ASFLAGS = -f macho64 -d LEADING_UNDERSCORE`

This allows it to compile (on two different macbook pros, one 2010 and the other 2016) without errors, but it crashes when I used the core in retroarch (I tested two games, also on both macs):

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [57321]

VM Regions Near 0:
--> 
    __TEXT                      1001a7000-1006a1000    [ 5096K] r-x/rwx SM=COW  /Users/*/RetroArch-gl.app/Contents/MacOS/RetroArch

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_platform.dylib      	0x00007fff2038818c _platform_memmove$VARIANT$Haswell + 268
1   mupen64plus_next_libretro.dylib	0x00000001219e70cd opengl::BufferedDrawer::_updateRectBuffer(graphics::Context::DrawRectParameters const&) + 653
```
I'm not sure if it was related to the change to GLFunctions.cpp, but I thought I would mention the error in case it was (and someone might be familiar with it..).

m4xw:
@aschain can u upload the mupen64plus_next_libretro.dylib you used in that log?
If you cant i need the matching ver to a crash report and i can tell you what blows up

m4xw:
Also was that with the define removed or changed to get proc?

aschain:
Here is the [mupen64plus_next_libretro.dylib](https://1drv.ms/u/s!Atqf0AIMAwothIcXVPW9UEZtZAJ2Gg?e=lePDDN).  And a [full error report](https://pastebin.com/mBfeFzj0).  I ran Super Mario 64 on the Parallel core, which sucessfully ran, and then switched to my compiled mupen64plus_next core, which crashes with that error. I was using the GL (obviously can't use the metal one) version of RetroArch 1.9.0 stable for os x downloaded from the website. 
This was with the define in GLFunctions.cpp removed with ifdef APPLE, but I also compiled by changing the line to:
```
#ifndef __APPLE__
	ASSIGN_PROC_ADR(PFNGLTEXIMAGE2DMULTISAMPLEPROC, TexImage2DMultisample);
#else
	GL_GET_PROC_ADR(PFNGLTEXIMAGE2DMULTISAMPLEPROC, TexImage2DMultisample);
#endif
```
It still compiles ok, but I get the same error.

m4xw:
@aschain GLideN64\src\Graphics\OpenGLContext\opengl_GLInfo.cpp line 116, set `bufferStorage = false;` and try again

aschain:
Same error.
I tried bufferStorage=true, and got a slightly different error:
```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [68704]

VM Regions Near 0:
--> 
    __TEXT                      102e77000-103371000    [ 5096K] r-x/rwx SM=COW  /Users/*/RetroArch-gl.app/Contents/MacOS/RetroArch

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   ???                           	000000000000000000 0 + 0
1   mupen64plus_next_libretro.dylib	0x0000000114ff9866 opengl::BufferedDrawer::BufferedDrawer(opengl::GLInfo const&, opengl::CachedVertexAttribArray*, opengl::CachedBindBuffer*) + 374



```

m4xw:
@aschain revert the change, go to GLideN64\src\Graphics\OpenGLContext\opengl_ContextImpl.cpp line ~70 and rip the if() and buffered drawer out and just leave Unbuffered
> m_graphicsDrawer.reset(new UnbufferedDrawer(m_glInfo, m_cachedFunctions->getCachedVertexAttribArray()));

Does your RA use GL? If so, which GL ver?

aschain:
opengl_ContextImpl.cpp change gave the same error I got the first time.
RA uses GL. There is also a metal2 version available from the website, but that one does not allow ParaLLel to run. 
I explicitly set the driver to GL, but the crash still happens. Does RA say what GL version it's using? In RA System Info, it says OpenGL- yes, OpenGLES-no. I know the mac I'm using defaults to GL2 but I think is capable of GL4 as well.
Oh, in RA logs, it says: Version: 2.1 ATI-4.0.47.
Hmm, I should probably have included the logs, [here](https://pastebin.com/tn7tiyTq) is one where I ran ParaLLel first, and then mupen64plus_next after.
Interestingly, it seems to crash right after 
`[libretro INFO] mupen64plus: Init new dynarec`
So maybe it is the NASM compile after all.

m4xw:
Was there a log created in system/Mupen64plus/ ?
You coul quickly rule that out by just changing the config to cached_interpreter manually

aschain:
OK, used cached interpreter (and pure interpreter) and I still get the same _updateBufferRect error.
No log in system/Mupen64plus, or logs/Mupen64Plus-Next.

aschain:
Oh, I just turned on the RA setting "Hardware Shared Context" and now it is working!

m4xw:
> 
> 
> Oh, I just turned on the RA setting "Hardware Shared Context" and now it is working!

It's spooky, because its supposed to only work without it.
Weird that mac requires it, but now I can look at manually merging those smaller changes and enforce it for the time being..
Thanks for testing.

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
ethteck:
```
GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp:345:2: error: assigning to
      'PFNGLTEXIMAGE2DMULTISAMPLEPROC'
      (aka 'void (*)(unsigned int, int, int, int, int, unsigned char)') from incompatible type 'void
      (GLenum, GLsizei, GLenum, GLsizei, GLsizei, GLboolean)' (aka 'void (unsigned int, int, unsigned
      int, int, int, unsigned char)'): type mismatch at 3rd parameter ('GLint' (aka 'int') vs 'GLenum'
      (aka 'unsigned int'))
        ASSIGN_PROC_ADR(PFNGLTEXIMAGE2DMULTISAMPLEPROC, TexImage2DMultisample);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        ```
        
  ^ if not changed

PR #258, opened 2020-09-08
Support aarch64 architecture for rpi4

dmrlawson:
- Set dynarec to aarch64
- Take out NEON references because they're implied by the aarch64 compiler.

I will update https://github.com/libretro/Lakka-LibreELEC/pull/1022 to not set `HAVE_NEON=1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,16 +153,22 @@ else ifneq (,$(findstring rpi,$(platform)))
       EGL_LIB := -lbrcmEGL
       INCFLAGS += -I/opt/vc/include -I/opt/vc/include/interface/vcos -I/opt/vc/include/interface/vcos/pthreads
    endif
-   WITH_DYNAREC=arm
    ifneq (,$(findstring rpi2,$(platform)))
-      CPUFLAGS += -mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard
-      HAVE_NEON = 1
+      CPUFLAGS += -mcpu=cortex-a7
+      ARM_CPUFLAGS = -mfpu=neon-vfpv4
    else ifneq (,$(findstring rpi3,$(platform)))
-      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard
-      HAVE_NEON = 1
+      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53
+      ARM_CPUFLAGS = -mfpu=neon-fp-armv8
    else ifneq (,$(findstring rpi4,$(platform)))
-      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard
+      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72
+      ARM_CPUFLAGS = -mfpu=neon-fp-armv8
+   endif
+   ifeq ($(ARCH), aarch64)
+      WITH_DYNAREC=aarch64
+   else
+      WITH_DYNAREC=arm
       HAVE_NEON = 1
+      CPUFLAGS += $(ARM_CPUFLAGS) -mfloat-abi=hard
    endif
    COREFLAGS += -DOS_LINUX
    ASFLAGS = -f elf -d ELF_TYPE


m4xw:
@dankcushions Wanna review?

joolswills:
It may not technically be correct to use HAVE_NEON - depending if it's used to include any neon asm etc. A32 Neon code won't work on aarch64. HAVE_NEON is usually used in libretro to signify 32bit neon support or at least from what I've seen.

on another topic - I don't think there is any performance benefit from using auto vectorisation from my own tests on armv7/armv8.

dmrlawson:
> It may not technically be correct to use HAVE_NEON - depending if it's used to include any neon asm etc. A32 Neon code won't work on aarch64. HAVE_NEON is usually used in libretro to signify 32bit neon support or at least from what I've seen.
>
> on another topic - I don't think there is any performance benefit from using auto vectorisation from my own tests on armv7/armv8.

It was suggested to keep `HAVE_NEON=1` to ensure we get the optimisations. I had to remove `-D__ARM_NEON__` for exactly the reason you describe though (libretro-common uses some A32 instructions when it's defined).

I haven't done any formal performance comparison so I can't speak to that. I was just aiming to include the same options where possible.

m4xw:
> on another topic - I don't think there is any performance benefit from using auto vectorisation from my own tests on armv7/armv8.

It's night and day for Switch.
If you use clang tho, then yes there will be hardly a difference
Also yes this should be technically handled by system headers, but well the fun of homebrew toolchains

m4xw:
Also the neon flag is supposed to set these https://github.com/libretro/mupen64plus-libretro-nx/blob/882719706aee4c8e470af90abd950b1c2420c8d6/GLideN64/src/gSP.cpp#L489
https://github.com/libretro/mupen64plus-libretro-nx/commit/f09c7b469aa0ab6aa10f1b20eeacc8c266950561
https://github.com/libretro/mupen64plus-libretro-nx/blob/thr_neon/GLideN64/src/Neon/RSP_LoadMatrixNeon.cpp
Which are all using intrinsics

dankcushions:
> It's night and day for Switch.
If you use clang tho, then yes there will be hardly a difference

gcc, here. any good benchmarks to test this?

dmrlawson:
I should point out, I don't think this builds as is now. I'm sure it did, but I can't reproduce it. With `HAVE_NEON=0` or unset it was fine. I think it's trying to build with the `arm` dynarec

dmrlawson:
It's working now with the optimisations from the switch config. I can squash the commits later but how do you feel it at the moment?

m4xw:
> 
> 
> It's working now with the optimisations from the switch config. I can squash the commits later but how do you feel it at the moment?

Does... busy.. count? :shrek:

dmrlawson:
> Does... busy.. count? :shrek:

Busy is a feeling :P

I think that this won't affect anything unless `ARCH=aarch64` and platform contains rpi4, so even if it's not an ideal configuration for now it won't affect anything else.

dankcushions:
i don't think i really get adding the switch flags - i doubt they are all appropriate. i think unless there's some kind of argument/benchmark for each flag you should keep it simple. things like
```
-fno-rtti
-std=gnu++11
-DOS_LINUX
-DEGL
```
don't appear to be anything to do with aarch64/pi4, right?

dmrlawson:
> i don't think i really get adding the switch flags - i doubt they are all appropriate. i think unless there's some kind of argument/benchmark for each flag you should keep it simple. things like
> 
> ```
> -fno-rtti
> -std=gnu++11
> -DOS_LINUX
> -DEGL
> ```
> 
> don't appear to be anything to do with aarch64/pi4, right?

You're probably right - I'm not very familiar with gcc options. I'll remove them.

dmrlawson:
Having said that.. GLideN64 appears to use `OS_LINUX` and `EGL`. I'm not sure what EGL does (is it related to full OpenGL?), but I think `OS_LINUX` might be useful.

dankcushions:
👍 
```
PLATCFLAGS = -Ofast -funsafe-math-optimizations -fPIE -ffunction-sections -fdata-sections -ftls-model=local-exec
PLATCFLAGS += $(INCLUDE) -D_GLIBCXX_USE_C99_MATH_TR1 -D_LDBL_EQ_DBL -funroll-loops
```
-ffunction-sections
-fdata-sections:
> Only use these options when there are significant benefits from doing so. When you specify these options, the assembler and linker create larger object and executable files and are also slower. These options affect code generation. They prevent optimizations by the compiler and assembler using relative locations inside a translation unit since the locations are unknown until link time. An example of such an optimization is relaxing calls to short call instructions.
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

ftls-model=local-exec
not sure about this one...

GLIBCXX_USE_C99_MATH_TR1
D_LDBL_EQ_DBL
again these seem like they don't have anything to do with aarch64, maybe library things? if it compiles without, then surely not needed.

funroll-loops
not to do with aarch64. none of the opts above have to do with aarch64 really! why aren't they just defaults for the makefile if they're desired?

dankcushions:
> Having said that.. GLideN64 appears to use `OS_LINUX` and `EGL`. I'm not sure what EGL does (is it related to full OpenGL?), but I think `OS_LINUX` might be useful.

sure but aarch64 doesn't change the linux or gl requirements, right? so if it's needed, it's needed under a different if. i suspect it's not needed as everything seems to work fine without, but either way.

dmrlawson:
tls is thread local storage: https://scc.ustc.edu.cn/zlsc/sugon/intel/compiler_c/main_cls/copts/ccpp_options/option_ftls_model.htm

I agree, none of this is aarch64 specific and should perhaps be set globally. For now I'd just like it to compile and not break anything else so I'll simplify it

dankcushions:
thanks!

it's still kind of wrong as pi3 can also be aarch64. i'd have it as:

```
   ifneq (,$(findstring rpi2,$(platform)))
      CPUFLAGS += -mcpu=cortex-a7 -mfpu=neon-vfpv4
   else ifneq (,$(findstring rpi3,$(platform)))
      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8
   else ifneq (,$(findstring rpi4,$(platform)))
      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72 -mfpu=neon-fp-armv8
   endif
   ifeq ($(ARCH), aarch64)
      WITH_DYNAREC=aarch64
   else
      WITH_DYNAREC=arm
      HAVE_NEON = 1   
      CPUFLAGS += -mfloat-abi=hard
   endif
```
idk if the redundant -mfpu options will cause errors for aarch64. might have to if those...

dmrlawson:
Yeah that makes sense.

I'll double check tomorrow but from memory the -mfpu options do cause errors (my understanding is that neon is implied by aarch64).

dmrlawson:
That looks a lot neater, thanks for the suggestion.

I seem to remember this being true of the `-mfpu` option when I tried it before: https://stackoverflow.com/questions/29851128/gcc-arm64-aarch64-unrecognized-command-line-option-mfpu-neon so I'm not going to worry about that.

rpi2 probably won't work, is that an issue? As I understand it only certain revisions support 64 bit so it's a bit complicated anyway.

We compile with `-O3` which turns on `-ftree-vectorize` which to my understanding will generate NEON instructions where possible.

Let me know if you're happy with this and I'll squash the commits

dankcushions:
just need to remove the -mfloat-abi=hard from rpi2 and otherwise it's fine :) thanks for this

dmrlawson:
Ah yes, good point.

I think `mfpu=neon-fp-armv8` will fail on rpi2 also

dankcushions:
maybe instead something like   

```
 ifneq (,$(findstring rpi2,$(platform)))
      CPUFLAGS += -mcpu=cortex-a7 -mfpu=neon-vfpv4
   else ifneq (,$(findstring rpi3,$(platform)))
      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a53 -mfpu=neon-fp-armv8
   else ifneq (,$(findstring rpi4,$(platform)))
      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72 -mfpu=neon-fp-armv8
   endif
   ifeq ($(ARCH), aarch64)
      WITH_DYNAREC=aarch64
      MFPU_BLACKLIST= -mfpu=neon-vfpv4 -mfpu=neon-fp-armv8
      CPUFLAGS:=$(filter-out $(MFPU_BLACKLIST),$(CPUFLAGS))
   else
      WITH_DYNAREC=arm
      HAVE_NEON = 1   
      CPUFLAGS += -mfloat-abi=hard
   endif
```
guess it gets a little complicated...

dmrlawson:
@dankcushions I modified your suggestion a bit to only add the extra bits in the 32bit case, rather than filtering things out.

What do you think?

dankcushions:
looks good to me, and more readable!

dmrlawson:
Squashed the commits - ready to merge?

PR #259, opened 2020-09-16
Ultrawide support

AngheloAlf:
Hi!

This change adds support for 64:27 aspect ratio, oftenly labeled as 21:9 as a marketing term.

Add the `64:27 adjusted` option to the `Aspect Ratio` in-menu option, and adds a `64:27 Resolution` in-menu option to let the user select the resolution for this aspect ratio.

Some comparison screenshots:

| Aspect ratio | Title screen screenshot | Lon lon ranch screenshot |
| - | - | - |
| 4:3 | <a href="https://postimg.cc/tZyKZjK9" target="_blank">titlescreen1-4-3</a> | <a href="https://postimg.cc/qtfVrnQh" target="_blank">lonlonranch-4-3</a> |
| 16:9 | <a href="https://postimg.cc/tsxwBJyX" target="_blank">titlescreen1-16-9</a> | <a href="https://postimg.cc/zyy88qrM" target="_blank">lonlonranch-16-9</a> |
| 64:27 | <a href="https://postimg.cc/9zgkVZQm" target="_blank">titlescreen1-64-27</a> | <a href="https://postimg.cc/G4Tw6zPn" target="_blank">lonlonranch-64-27</a> |

More comparison screenshots can be seen [here](https://postimg.cc/gallery/hx9zyP8).

Tested on RetroArch 1.9.0, Pop!_OS 20.04, using a Zelda OoT randomized rom.


m4xw:
I dont like this PR (and its not your fault).
This will need to be implemented differently but I am not entirely sure how I want it either, so give me a few days to think about it!
I am def. open for adding this - so keep the PR open for reference

m4xw:
Also btw a change like 043c67927f7a9a1e9859970562fc21f39488a9ee would need to happen upstream

AngheloAlf:
I also think that this should be implemented differently, but I tried make the change respecting the way it was already implemented.

> Also btw a change like [043c679](https://github.com/libretro/mupen64plus-libretro-nx/commit/043c67927f7a9a1e9859970562fc21f39488a9ee) would need to happen upstream

Oh, you're right, I'll see what can I do there.

m4xw:
> I also think that this should be implemented differently, but I tried make the change respecting the way it was already implemented.

Ye, I noticed that ;)
My pain limit for resolution selectors is 2 tho 🙈

AngheloAlf:
> My pain limit for resolution selectors is 2 tho :see_no_evil:

And what about just leaving a "4:3 Resolution" selector and a `Widescreen Resolution` selector?

Screenshot from 2020-09-16 21-24-02

The `Widescreen Resolution` selector would have all the resolutions for 16:9 and 64:27. It could even include other weirds aspect ratios.

Screenshot from 2020-09-16 21-24-40

The `Aspect Ratio` selector would only have 3 options. The option `Widescreen adjusted`  would calculate the aspect ratio based on the selected `Widescreen resolution`. The other options would force the corresponding aspect ratio.

Screenshot from 2020-09-16 21-25-05

m4xw:
I like that more, Will gather some opinions

m4xw:
image
We should keep backwards compat here for now, can remove it later after most of the users saved their config after the deployment

AngheloAlf:
Something like this? I undid the settings change, so it has the same options, but it still provides compatibility with wider screens.

m4xw:
Not sure about `,`'s, but I am no native speaker
image
Will check the rest when I find some time

AngheloAlf:
I think they are correct, but I can remove them if you prefer it.

m4xw:
> 
> 
> I think they are correct, but I can remove them if you prefer it.

Yea please do i think they ruin the pacing of the sentence.

AngheloAlf:
Done. Do you suggest any more changes?

libretro/libretro_core_options.h
```diff
@@ -114,28 +114,36 @@ struct retro_core_option_definition option_defs_us[] = {
     {
         CORE_NAME "-169screensize",
         "16:9 Resolution",
```
m4xw, line 116: Maybe turn it "Wide Resolutions"

libretro/libretro.c
```diff
@@ -1024,13 +1025,16 @@ static void update_variables(bool startup)
        {
           if (!strcmp(var.value, "16:9 adjusted")) {
              AspectRatio = 3;
-             retro_screen_aspect = 16.0 / 9.0;
+             /// `retro_screen_aspect` is calculated on the fly when retrieving the `-169screensize` setting.
```
m4xw, line 1028: no triple /

m4xw (changes requested):
This should be good to go otherwise, I will push it to a branch tomorrow and let a few ppl test. if all works out I will merge.
Not sure if i missed some behaviour.

libretro/libretro_core_options.h
AngheloAlf: Yeah, I like that.

I think it should be "Wide Resolution" (without "s") to be consistent with the "4:3 Resolution" option. The other option would be to change both ("4:3 Resolution" and "Wide Resolution")

m4xw: Sure

AngheloAlf:
Done both changed. Anything else?

libretro/libretro_core_options.h
```diff
@@ -113,29 +113,37 @@ struct retro_core_option_definition option_defs_us[] = {
     },
     {
         CORE_NAME "-169screensize",
```
m4xw, line 115: Now that I think about it, add a comment that we keep the key for compat.
Also sorry for stretching out this PR a little, just hesitant on stuff with configs ;)

AngheloAlf: No problem. It's understandable. Stuff needs to be done correctly.

AngheloAlf: Is this ok? Or should I change it?
image

m4xw: Keeping -> Keep, key name -> key.

After the dot add `// TODO: revisit-later` (including the redundant //)

AngheloAlf: Like this?
image

Anything else?

libretro/libretro.c
```diff
@@ -591,6 +591,7 @@ void update_controllers()
 static void update_variables(bool startup)
 {
     struct retro_variable var;
+    const char *screen_size_key = NULL;
```
m4xw, line 594: I think this might be unitialized on a 2nd not `startup` call (happens when core options are changed.. after start, but most wont re-apply), so `var.key` in line +1048 will request with `null`, but since we don't update it, it causes no side-effect.
Could you confirm this behaviour? We probably want to make it `static`, I dont like asking for some NULL key if thats the case.

m4xw (changes requested):
See topic

libretro/libretro.c
AngheloAlf: I don't know how to trigger the 2nd `startup` call. I tried messing with the core options when running, a lot of times, but weren't able to trigger this function again. Also tried stopping the core and opening again (with and w/o option changed) with no success. Also I have been using this branch a lot, and I haven't had any issues.

But, I'm with you about not requesting `NULL` keys, that can't go well. 

I'll make that variable `static`, and also assign to it "-43screensize" instead of `NULL`, so it will always request a valid key.

m4xw:
Did a gliden rebase that I am gonna push later and also added this PR on top of it on a branch https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/pipelines/2831
Can you check if those builds work out for you? I will have it tested rn as well

AngheloAlf:
> Can you check if those builds work out for you? I will have it tested rn as well

I tested the artifact in that page, and also I compiled and tested that code myself in my machine, just to be sure. Both works without problems for me.

m4xw:
@AngheloAlf a bug i found, if you select 16:9 aspect (non adjusted) you dont calculate the aspect ratio leading to a 4:3 like image output
grafik
This is directly dependent.
For now we can change the display name of the options without  changing the config value (will be confusing but i eventually write about it in a blog post)

https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/blob/develop/libretro/libretro_core_options.h#L134-139
the null's -> "Widescreen", "Widescreen (Adjusted)", leave 4:3, dont touch the left value.

All the settings I have hooked up under the hood so u just need to set it right (as you found out ;-))
So we can just calc it for 2 || 3, I know stretched isnt so insteresting for ultra wide, but I might just add a few other odd ones in the future, so i like to keep that open

m4xw:
Note, you also want to test the native res factor behaviour, aspect still applies

m4xw:
Another note, use cached interpreter if u want to play randomizer ;)

AngheloAlf:
> @AngheloAlf a bug i found, if you select 16:9 aspect (non adjusted) you dont calculate the aspect ratio leading to a 4:3 like image output
> grafik

That's weird, it should be calculated here:
image
Anyway, I'll look up what's the problem.

> This is directly dependent.
> For now we can change the display name of the options without changing the config value (will be confusing but i eventually write about it in a blog post)
> 
> https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/blob/develop/libretro/libretro_core_options.h#L134-139
> the null's -> "Widescreen", "Widescreen (Adjusted)", leave 4:3, dont touch the left value.

Yeah, you're right. I don't know how I didn't think about that. I'll change it.

> Note, you also want to test the native res factor behaviour, aspect still applies

I have been playing with this enabled (6x I think) with no problem, but I'll test it anyway and look up for some side effect.

> Another note, use cached interpreter if u want to play randomizer ;)

Oh. Why? I thought the cached interpreter were slower. I remember reading somewhere about changing the TLB exceptions option when playing rando, so I'll look up that option too. Thanks

m4xw:
> That's weird, it should be calculated here:

Thats 16:9 tho and its a ultrawide resolution ;)

Dont forget this is the "wide" option yet, so we need to calc it.

> Oh. Why? I thought the cached interpreter were slower. I remember reading somewhere about changing the TLB exceptions option when playing rando, so I'll look up that option too. Thanks

On PC? Shouldnt be significant and yea, because of the dynarec Issues, the TLB option is just a hack so you can basically save & restart (read: close content and start, otherwise it wont fix itself)

m4xw:
Btw for reference, this is the libretro aspect ratio.
The "AspectRatio" key (1, 2, 3) is set here in GLideN64 https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/custom/GLideN64/mupenplus/Config_mupenplus.cpp#L137
Its enforced here: https://github.com/libretro/mupen64plus-libretro-nx/blob/8a92a862b36bc89b3a7598aa2e45189371e47dac/GLideN64/src/DisplayWindow.cpp#L136

So my suggestion now is remove the "forced 16:9" (== 2) with "Stretch" (== 0), as seen here https://github.com/libretro/mupen64plus-libretro-nx/blob/8a92a862b36bc89b3a7598aa2e45189371e47dac/GLideN64/src/Config.h#L94-L100
Keep 4:3 what it is and then keep  how it is for "adjusted"

AngheloAlf:
> @AngheloAlf a bug i found, if you select 16:9 aspect (non adjusted) you dont calculate the aspect ratio leading to a 4:3 like image output
That's a weird bug. With that aspect selected, choosing a 16:9 resolution stretches it to 16:9 as expected, but selecting a wider one renders the game at 4:3.

For visual reference, this is how I see the game in "16:9" aspect ratio with "1920x1080" selected:
whole_screen_1920_1080
And this is how I see it with that same aspect ratio, but with "2560x1080" selected:
whole_screen_2560_1080

> Btw for reference, this is the libretro aspect ratio.
> The "AspectRatio" key (1, 2, 3) is set here in GLideN64 https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/custom/GLideN64/mupenplus/Config_mupenplus.cpp#L137
> Its enforced here:
> 
> https://github.com/libretro/mupen64plus-libretro-nx/blob/8a92a862b36bc89b3a7598aa2e45189371e47dac/GLideN64/src/DisplayWindow.cpp#L136
> 
> So my suggestion now is remove the "forced 16:9" (== 2) and replace with "Stretch" (== 0), as seen here
> 
> https://github.com/libretro/mupen64plus-libretro-nx/blob/8a92a862b36bc89b3a7598aa2e45189371e47dac/GLideN64/src/Config.h#L94-L100
> 
> Keep 4:3 what it is and then keep how it is for "adjusted"

That could explain the problem. I'll test that in a moment.

m4xw:
Also might as well call "16:9" now "Wide (Stretched)"

m4xw:
Also I guess it might be confusing with how it was named before for you, but 16:9 always meant set resolution, stretched and adjusted always means widescreen hack.
It was just a side effect of having seperate core options for it :P
If people really need such behaviour, i neet to add a "fit" mode

AngheloAlf:
You were right, now the stretched option works "as intended"
Screenshot from 2020-10-22 16-29-52
It's weird playing with the image so stretched, but maybe somebody likes it (?).

> Also might as well call "16:9" now "Wide (Stretched)"

Yeah, I was thinking something like that too. 
Also, what about renaming "4:3" to "Original (4:3)"?
image

AngheloAlf:
> Also I guess it might be confusing with how it was named before for you, but 16:9 always meant set resolution, stretched and adjusted always means widescreen hack.
> It was just a side effect of having seperate core options for it :P
> If people really need such behaviour, i need to add a "fit" mode

A bit 😅. I think the naming was fine if we only consider there were mainly two aspect ratios.

m4xw:
Scaling the aspect ratio further would be frontend business IMO. (Instead of "Core provided" in RA)
Thats why stretch exists, also for perf gains on 16:9

m4xw:
But why someone would use it with ultrawide... good question xD

m4xw:
https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/pipelines/2879/builds

AngheloAlf:
> https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/pipelines/2879/builds

Works with no issues for me (both artifact and compiling myself).

libretro/libretro.c
AngheloAlf: I'll mark this as resolved.

m4xw (approved):

m4xw:
There can be a few nits for documentation that i can think of, like annotating the types i quoted but this looks good to me and testing passes.
Ready to go?

AngheloAlf:
> There can be a few nits for documentation that i can think of, like annotating the types i quoted but this looks good to me and testing passes.
> Ready to go?

By annotating the types, you mean change some magic numbers for something less "magic"?

For example, use this enum:
https://github.com/libretro/mupen64plus-libretro-nx/blob/8a92a862b36bc89b3a7598aa2e45189371e47dac/GLideN64/src/Config.h#L94-L100
Like this:
image
(i know it wont work because it is c++ code, but that is the idea)

m4xw:
Yes, just `// Aspect::aStretch`, i do that somewhere else already..
Still hesitant to make libretro.c cpp :P

AngheloAlf:
> Yes, just `// Aspect::aStretch`, i do that somewhere else already iirc..

Like this?
image

> Still hesitant to make libretro.c cpp :P

I think its fine like it is right now.

m4xw:
Remove the Config::

AngheloAlf:
Ups. Fixed :sweat_smile:

m4xw (approved):
If this backfires, blame the beer :P

m4xw:
Thanks for this

AngheloAlf:
> Thanks for this

Thanks to you! 
It was a great team work.

Have a great day :3

m4xw:
@AngheloAlf I totally forgot, the stretch thing is to support widescreen via game patches ;)

AngheloAlf:
> @AngheloAlf I totally forgot, the stretch thing is to support widescreen via game patches ;)

Ohh, that make sense.
Now that you mention it, I think DK64 has widescreen support built-in, so the player had to stretch the image in the tv options. 
I wonder if other games have that option :thinking:

Jj0YzL5nvJ:
> I wonder if other games have that option too :thinking: 

http://www.gamepilgrimage.com/content/games-widescreen-support-standard-definition
https://youtu.be/eM16F_dM1Js
https://youtu.be/VA1H6dw6Fxo

PR #282, opened 2020-11-24
Restore Haiku build

kwyxz:
Looks like these disappeared during a force push. This restores the Haiku build.


m4xw:
Oops, probably forgot to exfiltrate the commit to my gliden repo

PR #284, opened 2020-12-02
libretro: Update gitlab ci extends ordering

webgeek1234:


m4xw:
This will be amended and cherry picked on develop.
Cursed merge commits.

PR #290, opened 2020-12-27
Add Jetson Xavier config

dmrlawson:
I made this for my own use but wondered if you'd mind it being merged. I suspect the user base for it will be pretty small but someone might find it useful.


m4xw:
Sure. This reminds me i need to fix this ifeq / ifneq findstring bs

PR #292, opened 2020-12-31
Disable Android X86 Dynarec

webgeek1234:
It is not PIC compliant and won't run on api 23+. NDK r22 uses a different flag to ignore shared textrels. But better to drop the dynarec entirely since the number of devices running <= api 23 and x86 userspace is pretty low at this point.


m4xw:
Obsolete.

PR #297, opened 2021-01-13
Add RPI4 64 + OGA

liberodark:
No change GLES
Sorry for revert my bad

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,9 @@ else ifneq (,$(findstring rpi,$(platform)))
    else ifneq (,$(findstring rpi4,$(platform)))
       CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72
       ARM_CPUFLAGS = -mfpu=neon-fp-armv8
+   else ifneq (,$(findstring rpi4_64,$(platform)))
+      CPUFLAGS += -mcpu=cortex-a72 -mtune=cortex-a72
+      ARM_CPUFLAGS =
    else ifneq (,$(findstring rpi,$(platform)))
       CPUFLAGS += -mcpu=arm1176jzf-s
       ARM_CPUFLAGS = -mfpu=vfp
@@ -258,6 +261,9 @@ else ifneq (,$(findstring odroid,$(platform)))
       else
          CPUFLAGS += -mcpu=cortex-a9 -mfpu=neon
       endif
+      # ODROIDGOA
+   else ifneq (,$(findstring ODROIDGOA,$(BOARD)))
+      CPUFLAGS += -march=armv8-a+crc -mfpu=neon-fp-armv8 -mcpu=cortex-a35 -mtune=cortex-a35
    else
       # ODROID-U2, -U3, -X & -X2
       CPUFLAGS += -mcpu=cortex-a9 -mfpu=neon


dmrlawson:
Rpi4 64bit is already supported, just set ARCH=aarch64.

Also I think this will hit the case above it anyway because it searches for rpi4 anywhere in the platform string.

Maybe line 171 could look at the platform string rather than ARCH?

liberodark:
It is better for auto build.
Because otherwise it requires going through a .patch

dmrlawson:
I thought it was neat to just have `platform=rpi4` and then setting `ARCH` to either `arm` or `aarch64`, since the cpu and supported features are the same either way, but I don't disagree that your approach is convenient (plus I think it matches what some other cores do and consistency is good!)

However this isn't going to do anything. The `rpi4` bit will be matched by the first clause here:

```
   else ifneq (,$(findstring rpi4,$(platform)))
      CPUFLAGS += -march=armv8-a+crc -mtune=cortex-a72
      ARM_CPUFLAGS = -mfpu=neon-fp-armv8
   else ifneq (,$(findstring rpi4_64,$(platform)))
      CPUFLAGS += -mcpu=cortex-a72 -mtune=cortex-a72
      ARM_CPUFLAGS =
```

so the second one won't have any affect. Also, we probably want the same optimisation for 32 and 64 bit, so if you've found setting `-mcpu=cortex-a72` is better we should do it for both.

You could remove that and change this bit to look for `rpi4_64` in `platform`:

```
   ifeq ($(ARCH), aarch64)
      WITH_DYNAREC=aarch64
      HAVE_NEON = 0
   else
      WITH_DYNAREC=arm
      CPUFLAGS += $(ARM_CPUFLAGS) -mfloat-abi=hard
   endif
```

We'd need to make sure Lakka was updated to use it like that.

rtissera:
The problem is ARM_CPUFLAGS = -mfpu=neon-fp-armv8 throwing an error in aarch64 target GCC I think.

liberodark:
@Ntemis can be look for lakka that is not a problem.

dmrlawson:
`ARM_CPUFLAGS` is only used here, and only in the `else` case of the code I quoted. 

If you change it to:

```
   ifneq (,$(findstring rpi4_64,$(platform)))
      WITH_DYNAREC=aarch64
      HAVE_NEON = 0
   else
      WITH_DYNAREC=arm
      CPUFLAGS += $(ARM_CPUFLAGS) -mfloat-abi=hard
   endif
```

it will do what you want, but will break the Lakka build until it's changed to set `platform=rpi4_64`

liberodark:
Done @Ntemis will make the changes for lakka.

dmrlawson:
That might work but it's not what I meant and still relies on `ARCH` being set. 

I suggest changing [this line](https://github.com/libretro/mupen64plus-libretro-nx/pull/297/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R178) _only_

liberodark:
Ok no problem done

m4xw:
> Sorry for revert my bad

It absolutely wasn't your mistake. It just got merged without my review.
It would've been reverted even if it was correct.
Sorry for getting you caught up in that :P

m4xw:
> 
> 
> The problem is ARM_CPUFLAGS = -mfpu=neon-fp-armv8 throwing an error in aarch64 target GCC I think.

It's default on for armv8 so its redundant and not required.
only might want "+crc" for the cpu opt

Can someone hit me up what the base Issue is? Iirc RPi64 works no issue with the unix target.
Is this for 64bit raspbian? Since the rpi target is intended for it.

m4xw:
also @dankcushions requesting your signoff on the change

liberodark:
Im re adding OGA  @m4xw 
And is not for armbian build.

dankcushions:
I'm afraid I don't like this part:

ARCH is set correctly when compiling on pi4 64-bit via https://github.com/libretro/mupen64plus-libretro-nx/blob/fa4dd669db306a750e6d595e56c0fe69d601f51b/Makefile#L62

if you remove the use of ARCH then it require all build scripts to use the pi4_64 platform. this would break current retropie, for example. retropie could be fixed to manually use WITH_DYNAREC=aarch64 i suppose, but i wonder what this change fixes?

liberodark:
Issue fixed.

m4xw:
> And is not for armbian build.

If its not raspbian, then it should not be using the rpi platform. It should already just work as-is.
Also what you added still doesnt work as it will still pick up the rpi4 and not rpi4_64 target

liberodark:
Ok im adding only OGA for make that simple.

PR #305, opened 2021-01-30
Makefile: added Mesa OpenGL ES opts for platform AML & RK / use correct libGLESv2 for standard builds

SupervisedThinking:
This issue https://github.com/libretro/mupen64plus-libretro-nx/issues/239 lead to https://github.com/libretro/mupen64plus-libretro-nx/pull/246 but this is only valid for some non-standard vendor blob packages e.g. https://github.com/EmuELEC/EmuELEC/blob/master/packages/graphics/opengl-meson/package.mk#L31-L33

According to libglvnd & mesa:
```
Khronos recommends that the GLES 3.1 library also be called libGLESv2.
It also requires that functions be statically linkable from that
library.
```

Sources:
https://www.khronos.org/registry/implementers_guide.html#idm58020557104
https://github.com/NVIDIA/libglvnd/issues/44
https://github.com/mesa3d/mesa/commit/5921f372c89a68fac6ddefc009442721d9df4db2

So my PR adds platform flags for mesa & mali builds and reverts to default `libGLESv2` for all other builds.

@shantigilbert have a look at these changes because after them you'll need platform flags for some of your builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,16 +282,23 @@ else ifneq (,$(findstring AMLG,$(platform)))
          CPUFLAGS += -mtune=cortex-a53
       endif
       GLES3 = 1
-      GL_LIB := -lGLESv3
    else ifneq (,$(findstring AMLGX,$(platform)))
       CPUFLAGS += -mtune=cortex-a53
       ifneq (,$(findstring AMLGXM,$(platform)))
          GLES3 = 1
-         GL_LIB := -lGLESv3
       else
          GLES = 1
-         GL_LIB := -lGLESv2
       endif
+   endif
+
+   ifneq (,$(findstring mesa,$(platform)))
+      COREFLAGS += -DEGL_NO_X11
+   endif
+
+   ifneq (,$(findstring mali,$(platform)))
+      GL_LIB := -lGLESv3
+   else
+      GL_LIB := -lGLESv2
    endif
   
    HAVE_NEON = 1
@@ -328,6 +335,10 @@ else ifneq (,$(findstring RK,$(platform)))
    else ifneq (,$(findstring RK3288,$(platform)))
       CPUFLAGS += -march=armv7ve -mtune=cortex-a17 -mfloat-abi=hard -mfpu=neon-vfpv4
       GLES3 = 1
+   endif
+
+   ifneq (,$(findstring mesa,$(platform)))
+      COREFLAGS += -DEGL_NO_X11
    endif
 
    GL_LIB := -lGLESv2


shantigilbert:
Funny you mentioned this, as I have been building internally with glesv2 from some other issues I was having, so thank you for the changes, I will test this and report back if needed.

shantigilbert:
No issues for me on this PR. But I do have to say I do not use the latest commits, I use b785150465048fa88f812e23462f318e66af0be0 since after that it became very slow for the low end devices I support on aarch64, but I did test the PR on latest.

m4xw:
> since after that it became very slow for the low end devices I support on aarch64

It absolutely should not. In fact it should be faster than ever, so citiation needed.

m4xw:
lgtm

m4xw:
> 
> 
> No issues for me on this PR. But I do have to say I do not use the latest commits, I use [b785150](https://github.com/libretro/mupen64plus-libretro-nx/commit/b785150465048fa88f812e23462f318e66af0be0) since after that it became very slow for the low end devices I support on aarch64, but I did test the PR on latest.

@shantigilbert 
Thinking of, this might be related to using Buffer swap on vertical interrupt. But this will likely cause other issues.
Fwiw that change should be a lot mitigated with  the threaded renderer enabled.
You can do a test with forcing Poll behaviour early in RA and change https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/custom/GLideN64/mupenplus/Config_mupenplus.cpp#L154 to `config.frameBufferEmulation.bufferSwapMode = Config::bsOnColorImageChange;`

shantigilbert:
I've been a bit busy that is why I didn't dig in finding what the real issue is, hence why I reverted back to an older version, I meant to open an issue soon, but I will try your suggestion and see if that fixes it. Also I am fine with using patches, specially if this is something that only affects me :) 

Here is an example on what I mean, on Mario Tennis, using b785150 it's smooth and you can play with (almost) no issues, using latest, every time you serve the ball (where it shows all these particles when hitting the ball) the emulation comes to a crawl, I am talking like (depending on the device) 0fps - 5 fps, then it picks up, but each time a player hits the ball,fps drop considerably.

shantigilbert:
Even with the patch it is still the same more or less: here is a small video comparison I made: https://youtu.be/qdy1ML0wuSA

This is running on a N2+ (stock clock) all settings default, I made sure to clean up any other settings, saves, etc, left by each version. 

On a lower end device (say a S905x2) the FPS drop even more than this.

And sorry for hijacking this PR, if you want I can open a proper issue for this.

m4xw:
The defaults need to be changed for low end ofc.
There have been some relevant changes but i would need to check when they happened
 > And sorry for hijacking this PR, if you want I can open a proper issue for this.
go ahead

PR #320, opened 2021-03-06
Fix for SW Ep1 Racer audio with other core timings.

mountnside:
The "special" interrupt is used as a major timing hack.

This fix removes that hack. It also has the side effect of fixing SW Ep1 audio in other counter factors than 1. It may have other effects in other games, which haven't been tested yet.

The DK64 vine problem (and many others) will require a much more thorough looking through. The interrupt system still seems to be a complete riddle to me in how it does things.

### DIFF
--- a/mupen64plus-core/src/device/r4300/interrupt.c
+++ b/mupen64plus-core/src/device/r4300/interrupt.c
@@ -263,6 +263,10 @@ void remove_event(struct interrupt_queue* q, int type)
     }
 }
 
+
+
+
+
 void translate_event_queue(struct cp0* cp0, unsigned int base)
 {
     struct node* e;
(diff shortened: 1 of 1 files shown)


m4xw:
https://github.com/mupen64plus/mupen64plus-core/pull/708
https://github.com/mupen64plus/mupen64plus-core/pull/701#issuecomment-586724230
To give you some background. I will check back with that on Gillou

mountnside:
Ideally then it would wrap to complete 2^32 I guess (or even could ditch the x86 recomp to allow for more register allocation for proper COUNT timing.)

I intend though to work on the Perfect Dark audio timing also. I am cleaning up those.

m4xw:
Whats the status quo?

mudlord:
I honestly would rather I forget I ever did any work on this, let alone **anything publically** to begin with. 

That's quite honestly the status quo.

pet-sematary

PR #334, opened 2021-04-22
Add ANGLE support to the upstream Mupen Libretro core

Nun-z:
This was done by JustMeDaFaq (https://github.com/JustMeDaFaq) and makes it easier to build the ANGLE version of the core for Xbox.  The core built from this source has been tested and it also works with Angrylion.

### DIFF
--- a/GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
+++ b/GLideN64/src/Graphics/OpenGLContext/GLFunctions.h
@@ -54,7 +54,9 @@ typedef void (APIENTRYP PFNGLDELETETEXTURESPROC) (GLsizei n, const GLuint *textu
 typedef void (APIENTRYP PFNGLGENTEXTURESPROC) (GLsizei n, GLuint *textures);
 typedef void (APIENTRYP PFNGLCOPYTEXIMAGE2DPROC) (GLenum target, GLint level, GLenum internalformat, GLint x, GLint y, GLsizei width, GLsizei height, GLint border);
 #endif
-
+#ifdef HAVE_ANGLE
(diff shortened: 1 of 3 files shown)


Nun-z:
@m4xw Is there any reason why this was closed?

m4xw:
First, don't just PR other ppl's stuff.
Second, the PR is incomplete. I already have a working angle integration which this must base on.
The ini drop of mupen ini fails, count per op wont set and many games will have troubles saving and/or running.
Also the missing addition of the vectored exception handler for the outputdebugstring logs is still a questionmark.
Technically its a frontend issue but it seems to occur because some tls shenanigans between msvc and mingw
Also I wont accept anything that points to outside the directory on develop for the Makefile. Also any change should target my angel branch (tho rebase is due).
Also readme isn't a place to store build configuration. Makefile is self documenting.

grafik
You also dropped a space.

Anyway, 3am. I am out

m4xw:
Or like i said on the other PR: That wont do.

Nun-z:
@m4xw It's not like I'm taking this from his repository.  He created the pull request for merging his code with my fork, which I accepted, and now I am passing that onto you.  I am in contact with him and we are collaborating.  If these changes are made, would the pull request be accepted?

1.  The fork will be changed to target the Angle testing branch.
2.  The ANGLE folder will be moved to the root of the repository.
3.  The changes to the readme will be removed.
4.  The code will be tested with a wide variety of games to ensure it is functional.

Nun-z:
This is posted in the wrong place.  Oops.

m4xw:
I wouldnt be against it. Ideally we move the angle folder to a env var (we may default to parent dir then with ?= as fallback).
Generally theres a bunch of stuff needed before angle_testing itself can be merged tho.
Your PR would be a first step however.
After we get the changes in, I can rebase it for you.

Nun-z:
@m4xw Can you please take a look at this commit?  The environment variable was added, and it defaults to the parent directory.

PR #336, opened 2021-04-29
Fix OSX with Angrylion

kivutar:
Not a final version. But we can start discussing some fixes here.


m4xw:
Because of your absence on the last PR, OSX support comes in through https://github.com/libretro/mupen64plus-libretro-nx/pull/286/files which is in staging.

m4xw:
I will consult with ethan

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
```diff
@@ -342,7 +342,9 @@ extern "C" void initGLFunctions()
 	ASSIGN_PROC_ADR(PFNGLDELETEFRAMEBUFFERSPROC, DeleteFramebuffers);
 	ASSIGN_PROC_ADR(PFNGLGENFRAMEBUFFERSPROC, GenFramebuffers);
 	ASSIGN_PROC_ADR(PFNGLFRAMEBUFFERTEXTURE2DPROC, FramebufferTexture2D);
+#ifndef __APPLE__
```
m4xw, line 345: Theres a type conflict and we fixed it on the gdb pr, which seemed to work.
need to check where its used and if its a downside, otherwise i'd suggest ifdef the prototype.
Theres def a missmatch with some headers on apples side & khronos

Makefile
```diff
@@ -351,9 +351,12 @@ else ifneq (,$(findstring RK,$(platform)))
 else ifneq (,$(findstring osx,$(platform)))
    TARGET := $(TARGET_NAME)_libretro.dylib
    LDFLAGS += -dynamiclib
-   OSXVER = `sw_vers -productVersion | cut -d. -f 2`
+
+   OSXVER = $(shell sw_vers -productVersion | cut -d. -f 2)
    OSX_LT_MAVERICKS = `(( $(OSXVER) <= 9)) && echo "YES"`
-        LDFLAGS += -mmacosx-version-min=10.7
+   ifeq ($(OSX_LT_MAVERICKS),YES)
```
m4xw, line 357: can we just ifeq(,1)?

custom/GLideN64/mupenplus/CommonAPIImpl_mupenplus.cpp
```diff
@@ -94,9 +98,10 @@ void PluginAPI::FindPluginPath(wchar_t * _strPath)
 		_getWSPath(path, _strPath);
 	}
 #elif defined(OS_MAC_OS_X)
+	#define MAXPATHLEN 256
```
m4xw, line 101: I dont like this one. I could offer moving it to the mupen next shared header and we call it MAX_PATH like the other platforms?

m4xw (changes requested):
See comments

custom/GLideN64/mupenplus/CommonAPIImpl_mupenplus.cpp
kivutar: This one is a direct copy of what you find here https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/GLideN64/src/mupenplus/CommonAPIImpl_mupenplus.cpp#L110

Makefile
kivutar: This one was copied from the makefile of snes9x, which is knows to work. Without it, we get this error at link time:
```
clang: error: invalid deployment target for -stdlib=libc++ (requires OS X 10.7 or later)
```

GLideN64/src/Graphics/OpenGLContext/GLFunctions.cpp
kivutar: Oh, that's indeed way better.
I updated this PR with what ethan did on the types.
I confirm that it's compiling fine now.

kivutar:
OK I updated this PR with the change on the types.

The changes I see in https://github.com/libretro/mupen64plus-libretro-nx/pull/286/files are not enough to compile the core on my machine.

Please note that this core segfaults in RetroArch on OSX.
The only place where I got it to work on OSX was in Ludo by forcing angrylion+cdx4.

custom/GLideN64/mupenplus/CommonAPIImpl_mupenplus.cpp
kivutar: OK I pushed a commit to use PATH_MAX_LENGTH

m4xw:
> 
> 
> OK I updated this PR with the change on the types.
> 
> The changes I see in https://github.com/libretro/mupen64plus-libretro-nx/pull/286/files are not enough to compile the core on my machine.
> 
> Please note that this core segfaults in RetroArch on OSX in `BufferedDrawer::_updateRectBuffer`.
> The only place where I got it to work on OSX was in Ludo by forcing angrylion+cdx4.

With all the last changes?
Ethan had it functional with Gliden iirc

m4xw:
We will need to ifdef the GL proto change btw, apparently apple is a snowflake and uses a different typedef than khronos states?

custom/GLideN64/mupenplus/CommonAPIImpl_mupenplus.cpp
```diff
@@ -94,9 +99,9 @@ void PluginAPI::FindPluginPath(wchar_t * _strPath)
 		_getWSPath(path, _strPath);
 	}
 #elif defined(OS_MAC_OS_X)
-	char path[MAXPATHLEN];
-	uint32_t pathLen = MAXPATHLEN * 2;
-	if (_NSGetExecutablePath(path, pathLen) == 0) {
+	char path[PATH_MAX_LENGTH];
+	uint32_t pathLen = PATH_MAX_LENGTH * 2;
```
m4xw, line 103: This one is bothersome. Is this a out pointer?
If its a max len then it could write double of path_max_length to the string however by design _NSGetExecutablePath should never write more than that anyway i assume?

libretro-common/glsm/glsm.c
```diff
@@ -2500,7 +2500,7 @@ void rglProgramBinary(GLuint program,
 
 void rglTexImage2DMultisample( 	GLenum target,
   	GLsizei samples,
-  	GLenum internalformat,
+  	GLint internalformat,
```
m4xw, line 2503: ifdef apple i guess

libretro-common/include/glsm/glsmsym.h
```diff
@@ -435,7 +435,7 @@ void rglTexImage3D(	GLenum target,
  	const GLvoid * data);
 void rglTexImage2DMultisample( 	GLenum target,
   	GLsizei samples,
-  	GLenum internalformat,
+  	GLint internalformat,
```
m4xw, line 438: same as the other change

m4xw (changes requested):
I also suggest trying the shared hw context switch.
Just toggle whatever its set rn for you in RA:

libretro-common/include/glsm/glsmsym.h
icthus56: I successfully built the mupen64 core library after incorporating your changes, but the core itself crashes Retroarch.

My toolchain is XCode 8.2, under El Capitan 10.11.6. Is my platform too old for GlideN64?

m4xw: Where does it crash?

custom/GLideN64/mupenplus/CommonAPIImpl_mupenplus.cpp
kivutar: A pointer to pathLen is passed yes. I don't know why the code used to multiply this by 2.
I can remove it and see what happen.

kivutar: Is it possible that 2 is the size of char here?

kivutar: Yeah this is a buffer size that we pass as a pointer, so to handle unicode we might need to *2, that's the only explanation I can find.

libretro-common/include/glsm/glsmsym.h
icthus56: I captured the log, see https://pastebin.com/KNfLe6kJ for details.

kivutar: OK, got it to run, it was HAVE_THR_AL missing to get angrylion. The HLE renderers still crash. Try adding HAVE_THR_AL and force LLE=1 and angrylion in the config.

icthus56: @kivutar Thanks for that, it runs here now after forcing angrylion.
However, performance seems much worse than the parallel64 plugin.
I assume this is because I cannot run GLideN64.

kivutar: Angrylion has always been slow, accurate but slow.

kivutar:
@m4xw I think we can merge this, what do you think?

It doesn't allow GL renderers, but at least allows Angrylion. And it doesn't break anything.

m4xw:
The core does not allow for compiling without gliden so I don't want to merge AL only support that still compiles gliden.
Its really the min standard. As is users would get crashes on start since you didnt adapt the default either and its already proven to work on osx at least, so find out where its crashing lol.

kivutar:
Oh I was thinking that it would simply default to angrylion on libretro frontends that don't support hardware renderers.

icthus56:
> > Oh I was thinking that it would simply default to angrylion on libretro frontends that don't support hardware renderers.
> 
> I wouldnt mind if we make gliden optional, but thats a bunch of work..

Does OSX currently work with GlideN64 (as opposed to AngryLion)?  I've only gotten crashes and assumed it was incompatible.

m4xw:
It used to with simliar minimal changes.
Someone needs to take a look...

m4xw:
> 
> 
> Any updates on this, or should I just fire up a VM?
> 
> EDIT: I incorporated these changes and while it successfully built, the core crashes on launch.

yea u need to default to angrylion

kivutar:
It looks like the hardware renderer if working now:

Screen Shot 2021-11-11 at 12 27 46

I didn't change anything though. So maybe something changed in RA?

Xenthio:
was this ever finished?

i-pwl:
Yeah id love to know as well

m4xw:
closed in favour of #474

PR #339, opened 2021-05-21
Add sys/types.h for u_int declaration

markand:
On musl libc which is strictier POSIX compliant than glibc, the u_int types are not available (as specified by POSIX) unless you include sys/types.h

This commit fix the build on musl, and since it's mandatory remove it so Apple builds get it by default too.


m4xw:
Move it to line 27 and add a \n afterwards

m4xw:
We will find out if all platforms ship that header :P
https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/pipelines/26670

m4xw:
All passed 👍

Issue #219, opened 2020-06-23
iOS version black textures issues


WeedyWeedSmoker:
Hi @m4xw, I thought for a long time that the iOS 64-bit version always had some "black texture issues", but it seems that's only happening with the main iOS branch… https://forums.libretro.com/t/retroarch-1-8-5-certain-textures-are-black-mupen64plus-next/27308/9

This fork from @jet082 actually works fine : https://github.com/jet082/mupen64plus-libretro-nx/commit/a7adaf2e253a445081d5b450ef063f62e4dd5b8d

I really don't know how to help but the official .ipa release uses this commit and not the latest version from the main branch…

Thanks!

m4xw:
>Core version packed inside libretro official build is Mupen64 1.0 a7adaf2; core versione packed inside Weedy Weed Smoker ipa is 2.0.5 etc. It seems to me that both cores are broken in specifically different ways; let’s just say that “official” core 1.0 a7adaf2 seems a little less broken.

Kek, as far I am aware we don't ship it because one of the IOS bit targets fails on the bot.
Must be some stale build somewhere...

You should give rebase branch a shot, its due to be released very very soon, just pulled in my GLES fixes earlier

WeedyWeedSmoker:
@m4xw Yeah, I saw your rebase branch but didn't try to compile it yet at the time…

It fails with this error:
```
Undefined symbols for architecture arm64:
  "_using_tlb", referenced from:
      _virtual_to_physical_address in tlb.o
ld: symbol(s) not found for architecture arm64
clang: error
```

m4xw:
Ohh yea i forgot we don't have dynarec, 1 sec

m4xw:
https://github.com/libretro/mupen64plus-libretro-nx/commit/109b019ffb5f93c4dd5fad5bb4088556749b099c @WeedyWeedSmoker

WeedyWeedSmoker:
@m4xw Ok, it compiles fine, but it's still plagued by the black textures issue reported earlier…

m4xw:
Did you clear your shader cache?

WeedyWeedSmoker:
@m4xw Yeah, cleared both the cache folder and the shaders folder…

Here is what Yoshi's story looks like for example with @jet082's fork:
![IMG_0497](https://user-images.githubusercontent.com/26014958/85777124-892bca00-b74b-11ea-80fc-37e2c7eaffde.jpg)
[...]

PR #343, opened 2021-05-26
[iOS] Add iOS arm64 job to gitlab

yoshisuga:
Support building for iOS for the nightly builds


m4xw:
Deliberately disabled due to #219

PR #358, opened 2021-07-15
[GLN64] Add EnableTexCoordBounds core option

bslenul:
Example with Ogre Battle (left and bottom of the text box):

OFF:

image

ON:

image

I used the name and the description from https://github.com/libretro/mupen64plus-libretro-nx/blob/a3801217629d6affc6a7ffa4e620d84e22b4ee64/GLideN64/src/GLideNUI/configDialog.ui#L1550-L1553

image

Hope this is fine!

### DIFF
--- a/libretro/libretro_core_options.h
+++ b/libretro/libretro_core_options.h
@@ -362,6 +362,17 @@ struct retro_core_option_definition option_defs_us[] = {
         "Off"
     },
     {
+        CORE_NAME "-EnableTexCoordBounds",
+        "Enable native-res boundaries for texture coordinates",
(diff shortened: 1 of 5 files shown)


libretro/libretro_core_options.h
```diff
@@ -361,6 +361,17 @@ struct retro_core_option_definition option_defs_us[] = {
         },
         "Off"
     },
+    {
+        CORE_NAME "-EnableTexCoordBounds",
+        "Enable native-res bounds for texture coordinates",
```
m4xw, line 366: *boundaries

m4xw (changes requested):
change that,  leave the other text as is, will get too long otherwise

PR #368, opened 2021-08-16
Update core options

twinaphex:
NOTE: You will still need to manually update libretro-common to the latest version to get the updated header file required for this extra field.


twinaphex:
Hi @m4xw, please don't merge this anymore. This was causing too many problems and we're now reverting it forever.

PR #381, opened 2021-09-11
Added rpi3/4_64 to makefile

jdorigao:


m4xw:
@dankcushions

dankcushions:
tbh over at retropie we have long-avoided all the libretro core platform makefile opts, because they're often nonsense. for example, mtune is not needed when using mcpu, and really the micromanagement of march/mtune can be detrimental - just use mcpu. see https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu

that said, the -mfpu stuff is right, and the march/mtune stuff wasn't introduced in this PR.

libretro should manage all these platform-specific compiler opts centrally IMO. as it is there's just too many fires to fight.

m4xw:
I take that you consider this superseded by #397

Issue #383, opened 2021-09-18
[Parallel-RDP] Crashes on load for newer Intel Xe drivers - "This device probably does not support 8/16-bit storage"


retroNUC:
I say newer, but this stopped working a while back. I was previously using the separate ParaLLEl core with no issues until an Intel driver update in May, which broke rendering (showing "ERR" instead of a render picture). Tried to get it working again today by switching to this core (which seems to be the actively maintained one), but is straight up crashing on load instead.

CPU: Intel Core i7-1165G7
GFX: Intel Xe Graphics
OS: Windows 10 21H1

Last working driver was 27.20.100.9466 Retail (April 2021)
Stopped working from 27.20.100.9616 Beta (May 2021) onwards

I had directly reported this to the ParaLLEl developer, but they seem to be inactive on the repo - https://github.com/Themaister/parallel-rdp/issues/26

I also raised this to Intel directly, as this very likely seems to be a problem with their Vulkan implementation, but they won't look at the issue - https://community.intel.com/t5/Graphics/Vulkan-BETA-Drivers-Break-RetroArch-ParaLLel-RDP/m-p/1297881

If that is the case, could someone debug and provide the exact details about exactly what they broke in Vulkan that their developers can understand, so the Intel thread could be reopened and investigated properly? Thanks.

[retroarch__2021_09_18__09_52_56.log](https://github.com/libretro/mupen64plus-libretro-nx/files/7189627/retroarch__2021_09_18__09_52_56.log)

retroNUC:
Also attaching the output from vulkaninfo, showing that this graphics device does report as having the correct vulkan extensions that I think it's checking for:

```
VK_KHR_16bit_storage                      : extension revision 1
VK_KHR_8bit_storage                       : extension revision 1
```

[vulkaninfo.txt](https://github.com/libretro/mupen64plus-libretro-nx/files/7189648/vulkaninfo.txt)

m4xw:
Can you make sure you select vulkan in RetroArch driver settings and dont let it switch?

retroNUC:
Set both of those, still the same result. Errors in log look the same.

retroNUC:
Thanks for looking into it and tracking down the offending driver issue. I'll re-report the issue on the Intel community forums, as they seem a bit more willing to help when you can point directly at the issue with their API implementation.

retroNUC:
Tested as working in latest nightly core on Intel XE driver 30.0.100.9894. Thank you!

PR #388, opened 2021-10-10
[RFC] Fix Parallel-RDP on Intel Xe

stuken:
As noted in #383 ParaLLel-RDP crashes on current Intel Xe drivers. On newer drivers vkGetPhysicalDeviceMemoryProperties calls incorrectly populates the VkPhysicalDeviceMemoryProperties struct values meaning that all contained memoryTypes->propertyFlags fields return 1 instead of the or'd flag values.
This means that the find_memory_type method's lookups will never match based on the expected flags, the rdram buffer is not initialized and it crashes out.

This will force set the memoryTypes index to the appropriate value bypassing the lookup method if the driver vendor is Intel and the version is at least the current most one, 30.0.100.9862.


m4xw:
deprecated

PR #430, opened 2022-05-08
Update to modern AL+

mudlord:


m4xw:
Dunno how i missed this PR.
What is the upstream source so it can be turned subrepo

m4xw:
Going over at it reminds me that I need to deal with the do_work change, need a option to spin on it and a option to synchronize like now.
Background was the conditionals causing android to power down the cores / switch to low power cores on high end smartphones and causing a perf regression while the current way is more optimized for low end

mudlord:
is there much point to this? or should i close it?

m4xw:
leave it open for now

mudlord:
k

mudlord:
Untested, but added busywaiting of worker threads.

mudlord:
I need to relook at this. Seems the Android changes don't 100% affect things. Only just the thread waits, not thread creation.

PR #433, opened 2022-05-25
Use Fiber backend on Win32.

Themaister:
Works around driver crashes on AMD Windows since driver seems to rely on unwinds to work properly.
libco cannot support this without using a more robust backend.


m4xw:
Temporary merge, amend later

PR #439, opened 2022-07-15
update Makefile to enable GLES3 support for rpi4

asciibeats:


LibretroAdmin:
@m4xw This should be fine to merge unless there is a specific objection against it.

m4xw:
iirc there was some reason, cant recall tho

m4xw:
Since there seems to be zero interest, lets just try and see who complains

PR #446, opened 2022-07-30
Makefile fix to enable cross build for "unix-armv7-hardfloat-neon"

zoltanvb:
"armv" platform is recognized by the makefile, but cross-compilation on an x86_64 host was not successful for "unix-armv7-hardfloat-neon" (one of the options in libretro-super build scripts), as ARCH was detected from host. Added fixed override.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,8 @@ ifneq (,$(findstring unix,$(platform)))
    endif
 
    ifneq (,$(findstring armv,$(platform)))
+      ARCH = arm
+      WITH_DYNAREC = arm
       CPUFLAGS += -DARM -marm
       ifneq (,$(findstring cortexa8,$(platform)))
          CPUFLAGS += -mcpu=cortex-a8


m4xw:
Huh anyone even using that target anymore?
Super builds are deprecated since 2 years

zoltanvb:
It is still useful if someone wants to do an armv7 build that is less tied to a specific piece of hardware. Most cores (150+) can be built with this target.

m4xw:
I prefer to have it optimized for hardware whereever possible (especially on embedded)
Generally the target needs to be reworked to support armv8 etc too, but i guess its whatever rn

m4xw:
Btw if u build em like that u are support to put the WITH_DYNAREC on your make command line since its the bare uniform target. There can be armv7 devices with w^x enforcement on hardened system etc, so cant just enable dynarec for all, but given nobody uses it really, i dont have a strong opinion on that

zoltanvb:
I understand the drawbacks. Whether compiling cores for this target, and making them available, is useful or would only cause confusion and perception of bad performance, is something that could be discussed.

On the other side, this is 2 lines of change. Handling this target is already enabled in the makefile. First line is only making a change in cross-compilation case, And actually the situation is similar with the second line, WITH_DYNAREC would get the value "arm" anyway in line 67 if it were detected from uname.

m4xw:
Point is as far I consider it, those are all legacy targets and I dont really like adding stuff to things I don't maintain and were only carried over for parity, so consider it a unstable target.
Not against merging it until some refactor happens, currently branch is frozen waiting for rebase_gliden to be pulled so needs to wait.

PR #474, opened 2023-01-16
Build on OSX

warmenhoven:
This gets mupen64plus_next running on OSX, both on x64 and arm64.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,6 +28,12 @@ include:
   # Linux 64-bit
   - project: 'libretro-infrastructure/ci-templates'
     file: '/linux-x64.yml'
+  # MacOS
+  - project: 'libretro-infrastructure/ci-templates'
+    file: '/osx-x64.yml'
+  # MacOS ARM 64-bit
+  - project: 'libretro-infrastructure/ci-templates'
+    file: '/osx-arm64.yml'
   # Android    
   - project: 'libretro-infrastructure/ci-templates'
     file: '/android-jni.yml'
@@ -132,7 +138,7 @@ android-x86_64-gles3:
     - .libretro-android-jni-x86_64
     - .core-defs-gles3
 
-# Apple
+# iOS
 libretro-build-ios-arm64-gles3:
   variables:
     HAVE_DYNAREC: ""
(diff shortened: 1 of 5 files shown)


mupen64plus-rsp-paraLLEl/jit_allocator.cpp
```diff
@@ -31,7 +32,12 @@ Allocator::~Allocator()
 
 static size_t align_page(size_t offset)
 {
-	return (offset + 4095) & ~size_t(4095);
+#ifdef _WIN32
+    size_t pagesize = 4095;
+#else
+    size_t pagesize = sysconf(_SC_PAGESIZE) - 1;
```
m4xw, line 38: sysconf isn't available on all systems, for libnx i provide it myself from inside RA.
Can't rely on that for the core tho

m4xw (changes requested):

warmenhoven:
Something like this? https://github.com/Themaister/parallel-rsp/pull/5 (This is closer to what flycast does as well.)

m4xw:
I still need to do some local test, i think the ASSIGN_PROC_ADR change is sane to do on all platforms either way.
Other than that it looks acceptable, will get to it the next time i touch it (sometime this week)

warmenhoven:
Thanks! Yes, the `ASSIGN_PROC_ADR` change should be safe on all platforms. It could potentially mask problems, so I wasn't sure if it would be preferable to limit the cast to just Mac. I can change it so it applies to all platforms.

warmenhoven:
I've changed `ASSIGN_PROC_ADR` so it applies to all platforms instead of just osx, and I've rebased onto the current `develop` branch.

m4xw:
That reminds me, subrepo has a incorrect parent reference. Can u just squash out the parallel rsp change and i will just pull it on my end after merge?

warmenhoven:
Yep, I've removed the parallel-rsp change. The osx build might fail until the subrepo is pulled though as rsp_jit.cpp did have a cast fix.

m4xw:
I still see the mupen64plus-rsp-paraLLEl/jit_allocator.cpp change.
If its upstream then i dont need any change for the folder

warmenhoven:
Yep, sorry, missed that. Fixed now. Thanks!

m4xw:
Actually i forgot I dont have my own remote there.... subrepo was fine :P 
I'd prefer to not resolve that conflict in the future tho so will just force clone in the future but whatever for now

m4xw:
https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/jobs/3068798 arm64 is faulty

m4xw:
did a short change, must be arch detection being faulty for apple cross comp

PR #475, opened 2023-01-17
"correctTexrectCoords" isn't inconsistent anymore

bslenul:
Tetrisphere used to have `generalEmulation\correctTexrectCoords=2` in "GLideN64.custom.ini" which was incorrect, it was fixed a while ago (https://github.com/gonetz/GLideN64/pull/2702) and it now uses the proper `graphics2D\correctTexrectCoords=2` instead, since the .ini is now up-to-date on this repo we can remove that "inconsistent" part from the parser :)


m4xw:
Thanks, i didnt double check that one. Will get back to this

PR #477, opened 2023-01-27
Clean up properly on deinit to allow reinitialization

warmenhoven:
The two global variables we care about cleaning up properly, that are not currently, are `l_CoreInit` (only reset by `CoreShutdown`, which was not previously called from anywhere) and `l_ROMOpen`. `l_ROMOpen` is interesting because it is reset by `M64CMD_ROM_CLOSE` which is called from `retro_unload_game`; but sadly this is too early for it to be called, as `g_EmulatorRunning` is still `1`. So instead we have to wait for the emulator to have been stopped by `M64CMD_STOP`. Further, if the `current_rdp_type` is not `RDP_PLUGIN_GLIDEN64`, `M64CMD_ROM_CLOSE` is not invoked until `retro_deinit`.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -739,6 +739,10 @@ void retro_deinit(void)
        }
     }
 
+    CoreDoCommand(M64CMD_ROM_CLOSE, 0, NULL);
+
+    CoreShutdown();
+
     deinit_audio_libretro();
 
     if (perf_cb.perf_log)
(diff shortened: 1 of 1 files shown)


sergystepanov:
Hey. 
Any info on when this PR will be merged? I tested it with a custom Go frontend and it fixes a horrible up to 1GB memory leak each time a game is restarted.

m4xw:
Will be dealing with it properly when I rebase -core soon (next week)

m4xw:
Fixed in 90dae29e145f5ff07e753d602f2c40583e3afee7

sergystepanov:
Retested it and it appears to be working. Now we can run it on a cloud VPS, knowing it won't blow up. (;
@warmenhoven, @m4xw, thank you.

PR #482, opened 2023-03-11
Add tvos

warmenhoven:
tvOS is mostly the same as iOS (if you jailbreak you can run most iOS cores on tvOS), so this is mainly a Makefile change, leaving -DIOS and similar in place, and adding equivalent -DTVOS defines if necessary in the future (they aren't currently).


m4xw:
Didnt see, woops

PR #499, opened 2023-08-04
Fix loading roms larger than 64MB

Thundertheidiot:
This one character fix makes the core able to load roms that are larger than 64MB in size. For example Super Mario 64 romhacks made using the older tools can sometimes surpass 64MB, and mupen64plus_next isn't able to run them correctly without this fix.

### DIFF
--- a/mupen64plus-core/src/device/cart/cart_rom.c
+++ b/mupen64plus-core/src/device/cart/cart_rom.c
@@ -32,7 +32,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define CART_ROM_ADDR_MASK UINT32_C(0x03ffffff);
+#define CART_ROM_ADDR_MASK UINT32_C(0x0fffffff);
 
 
 void init_cart_rom(struct cart_rom* cart_rom,


m4xw:
I do have a hard time believing this is all thats required

Thundertheidiot:
Go ahead and try it out. This one romhack called b3313 surpasses 64MB on 0.9 and later versions, and this build fixes it.

soup6020:
https://github.com/mupen64plus/mupen64plus-core/issues/997
https://benedani.xyz/files/mupen/readme.html
more information about this, not sure who the original author of the latter page is though

Benedani:
> I dont need to try it, at best this will allow data access and maybe interpreter but i am positive code execution from that will fail

On N64, code isn't ever executed from the ROM. The only ROM accesses are DMA copies to RAM. This makes no change to the RAM layout, it only makes ROM addresses not wrap around at 64 MB.

> https://github.com/mupen64plus/mupen64plus-core/issues/997
> https://benedani.xyz/files/mupen/readme.html
> more information about this, not sure who the original author of the latter page is though

gee I wonder who it could be, if only there was something common in the url & the guy who made the issue or something

m4xw:
> On N64, code isn't ever executed from the ROM.

You can run code from uncached cart dom

PR #490, opened 2023-05-07
WIP - parallel rdp/rsp on iOS

warmenhoven:
This isn't ready to be merged yet because I played fast and loose a bit with the parallel-rsp subrepo; namely I pointed it at my fork temporarily to pick up the latest lightning JIT. Once that PR gets merged upstream I can redo the subrepo pull and update this PR.

PR #508, opened 2023-11-29
Enable angrylion on macos/ios/tvos

warmenhoven:


kerberjg:
I don't think it's working. Just installed the latest RetroArch Nightly (2024-01-06), with Mupen64 core 9dd7459 (same commit as above merge), and while the config says Angrylion is enabled (tried manually saving config and restarting app), it's still rendering using GlideN64.

For context, tvOS version is 17.2, I'm using the latest Apple TV model (A2843).

Let me know if there's anything I can do to help troubleshoot this

warmenhoven:
Hm, it's working on Mac and I have heard from one other person it's working on tvOS, but I'll double check again soon.

warmenhoven:
Oh, I see what is the matter. If the JIT isn't enabled it will force GLideN64.

warmenhoven:
Sorry, the earlier comment was with my other PR (https://github.com/libretro/mupen64plus-libretro-nx/pull/490) applied. I tried it again without that PR and yes, on iOS and tvOS it falls back to GLideN64 because only the HLE RSP plugin is available. So I've created another PR (https://github.com/libretro/mupen64plus-libretro-nx/pull/511) to enable the CXD4 LLE RSP plugin, which doesn't require a JIT.

kerberjg:
Nice, thanks for taking a look at it! 😊
I hope these get merged in soon ✨

PR #509, opened 2023-12-14
Add a button to quickly swap between Memory<->Rumble Paks

### LINKED ISSUE #502: controller pack hotkey
hello

i wonder if it would be possible to add an hotkey to swap  controller packs ( or at least from memory to rumble pack ) on the fly , cuz the n64 controller had not select button it would be a great place for it ^^

maybe i am missing something and its already possible and just can´t find it

anyways in some games its no problem to change it once but in  turok 2 as example you have to do it quite often ..an it breaks a bit the flow of the game

bslenul:
This PR adds a button to quickly swap from Memory Pak to Rumble Pak and vice-versa (RetroPad Select by default, or R3 if "Independent C-button Controls" is enabled):

https://github.com/libretro/mupen64plus-libretro-nx/assets/33353403/440b6d40-2350-4e46-b305-ae9f636e9569

Very useful for games compatible with both Paks, going back and forth in the Quick Menu every time you want to change quickly becomes annoying...

Just select "memory/rumble (swappable)" for any of the "Player N Pak" core options and it will be enabled for that controller. Also added a core option to select if the swap should be done on press or on hold (1, 2 or 3 seconds).

I hope this is fine, if it is then it closes #502

Also removed the `update_controllers()` call here: https://github.com/libretro/mupen64plus-libretro-nx/blob/f1ad37c3a9be64b499d9a36f57e41d59fb677c73/libretro/libretro.c#L1988-L1991 since it's already done at the end of `update_variables()` just above.

### DIFF
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -124,6 +124,10 @@ int d_cbutton;
 int u_cbutton;
 bool alternate_mapping;
 
+static bool swap_pak_enabled[4] = {false};
+static int  swap_pak_hold[4]    = {0};
+static int  swap_pak_delay      = 0;
+
 static uint8_t* game_data = NULL;
 static uint32_t game_size = 0;
 
@@ -764,6 +768,25 @@ void update_controllers()
         else if (!strcmp(pk1var.value, "transfer"))
             p1_pak = PLUGIN_TRANSFER_PAK;
 
+        // This is for the 'Memory/Rumble Pak Swap' button that lets you swap
+        // between the Memory Pak and the Rumble Pak on-the-fly.
+        // Initialize to Memory Pak, then it's handled by pak_swapping().
+        // As long as swap_pak_enabled is true we just keep the current value
+        // from controller[port].control->Plugin, or else the Pak would be
+        // reinitialized to Memory Pak every time core options are refreshed.
+        if (!strcmp(pk1var.value, "memory/rumble (swappable)"))
+        {
+            if (swap_pak_enabled[0])
+               p1_pak = controller[0].control->Plugin;
+            else
+            {
+               swap_pak_enabled[0] = true;
+               p1_pak = PLUGIN_MEMPAK;
+            }
+        }
+        else
+            swap_pak_enabled[0] = false;
+
         // If controller struct is not initialised yet, set pad_pak_types instead
         // which will be looked at when initialising the controllers.
         if (controller[0].control)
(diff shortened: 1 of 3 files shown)


libretro/libretro.c
```diff
@@ -1731,6 +1843,20 @@ static void update_variables(bool startup)
...
+    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
+    {
+       if (!strcmp(var.value, "no delay"))
+         swap_pak_delay = 0;
+       else if (!strcmp(var.value, "1s hold"))
+         swap_pak_delay = 60;
+       else if (!strcmp(var.value, "2s hold"))
+         swap_pak_delay = 120;
+       else if (!strcmp(var.value, "3s hold"))
+         swap_pak_delay = 180;
```
m4xw, line 1857: 1. The heck is that unit? Frames?
2. If its Frames/VI, then this is not correct depending on Buffer Swap mode for GlideN & PAL/NTSC configurations
3. atoi or sscanf, prefer not to have val and string mixed tho

m4xw (changes requested):

libretro/libretro.c
bslenul: Yeah it was supposed to be frames, I assumed 60fps and completely forgot about PAL 😓 
A bit annoying since `ROM_PARAMS` isn't initialized yet on the first `update_variables()` pass so it returns 0 (NTSC) as region even with a PAL game...

Would that be OK as a workaround (with libretro_core_options.h values adjusted to just "0/1/2/3")?
```cpp
    var.key = CORE_NAME "-pak-swap-delay";
    var.value = NULL;
    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
    {
       // If we're on the 1st pass we don't know the region of the game yet,
       // so we just keep the seconds for now, and we'll multiply by FPS
       // just a bit later in retro_load_game()
       swap_pak_delay = atoi(var.value);
       if (!startup)
          swap_pak_delay *= retro_get_region() == RETRO_REGION_PAL ? 50 : 60;
    }
```
then in retro_load_game() after we got the ROM infos:
```cpp
    // Now that we now the region, we can adjust swap_pak_delay
    swap_pak_delay *= retro_get_region() == RETRO_REGION_PAL ? 50 : 60;
```

As for the buffer swap I'm not sure how to handle that tbh 🤔 

A bit of a bummer, looks more complicated than I expected for my *very* limited C/C++ skills :(

bslenul: Changes pushed, AFAICT it works fine with the correct hold duration for NTSC/PAL now.

Still not sure to understand your "buffer swap" comment, doesn't retro_run run 60 (or 50 for PAL) times per second no matter what? Because basically the only thing I'm doing is increasing the frame count for the duration of the button being held once per retro_run.

bslenul: Different approach using `clock()` from time.h: https://github.com/bslenul/mupen64plus-libretro-nx/commit/b2d9c20ace3e8bf1bd28ce9d5a2419e60253686c#diff-0033c4d206c0f5ecb21459b1c42f39055aacfc7468b9729527c5dd94489c58aa

Otherwise I could simply drop the hold button feature and make the swap on button press only. I initially added the hold options to avoid misclicks, but if you think it's not worth I don't mind removing it.

m4xw: I will think of smth

PR #523, opened 2024-02-23
Enable parallel-rdp on iOS/tvOS

warmenhoven:


kerberjg:
@warmenhoven So happy to see this merged! 🥳
I'm gonna test it on my Apple TV soon, but perhaps do you know of any way of enabling JIT on tvOS?

warmenhoven:
If you're running pre-17.0 then the easiest way is with AltServer. As long as you have AltServer running somewhere on your network, RetroArch will use Bonjour to find it and automatically enable JIT.

If you're running 17.0+ then there's not currently a great way other than running in the Xcode debugger.

PR #526, opened 2024-03-19
Enable compilation on Fedora 41 on aarch64

pstef:
This gets rid of two compilation errors I was getting with the bundled zlib and minizip libraries, and one compilation error in libretro.c.


m4xw:
Hi, sorry missed this PR. I will check it out by next week. I could swear i had some of the type missmatch stuff fixed before i cant remember where it went...
Is this for you personal or for distribution on the package manager?
I think the zip libs are fair to use the distribution but just for the record i highly recommend using the bundled libpng, historically that caused too much issues.

pstef:
I did this for my personal use and I thought I'd share.

pstef:
This is starting to conflict with other people's commits.

m4xw:
> This is starting to conflict with other people's commits.

Well yours already conflicts with my changes too ;)

Makefile.common
```diff
@@ -175,6 +176,9 @@ ifeq ($(SYSTEM_ZLIB), 1)
 	CFLAGS += $(shell pkg-config --cflags zlib)
 	LDFLAGS += $(shell pkg-config --libs zlib)
 else
+ifneq (,$(findstring unix,$(platform)))
+	PLATCFLAGS += -DHAVE_UNISTD_H
```
m4xw, line 180: Why is this required?

pstef: Without it `custom/dependencies/libzlib/zconf.h` doesn't include unistd.h and compilation fails on unknown functions read(), close(), write().

I don't insist on this solution, but something is needed to remedy this issue.

m4xw: Hm how didnt this blow before? Did it link implicitly? Not sure if large file support would be more appropriate.

pstef: No idea. It's not something I have tested previous versions against.

pstef: I just compiled this core on a Debian Trixie, which is a testing release but not as experimental as Fedora Rawhide is. The GCC version here is 13 as opposed to GCC 14 in Rawhide. The implicit declarations are reported by GCC 13 as well, but only as warnings. Most likely this warning has been promoted to an error on somewhere the way to GCC 14.1.

m4xw:
Can you try if fpermissive in cflags also works?

Makefile.common
m4xw: https://gcc.gnu.org/gcc-14/porting_to.html meh

pstef:
I'm already using this patch and whether you take it or not will have to be your choice and responsibility, not mine.

m4xw:
Btw i have been reproducing this with emscripten so will get that in after my update round

PR #541, opened 2024-06-22
Parallel RSP pagesize at build time

jonian:
This adds support for defining pagesize at buildtime for parallel RSP. Relevant issues on RPi5 from other libretro cores.

https://github.com/flyinghead/flycast/issues/1288
https://github.com/realnc/dosbox-core/issues/57
https://github.com/schellingb/dosbox-pure/issues/488

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,10 @@ ifeq (,$(ARCH))
    ARCH = $(shell uname -m)
 endif
 
+ifeq ($(shell uname -m),aarch64)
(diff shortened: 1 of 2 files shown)


m4xw:
Note: Fix for Dynarec

Adjust alignment at https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/mupen64plus-core/src/device/r4300/r4300_core.h#L184
Precompiled asm_defines_gas and asm_defines_nasm need to be deleted out of the source tree
Theres a issue with order of include dirs and it seems to pick the static ones over them

PR #545, opened 2024-07-23
Add Memory Map Setup

JoshuaEagles:
This PR adds a memory mapping for RDRAM, so that things like the network commands for `READ_CORE_MEMORY` and `WRITE_CORE_MEMORY` work for N64 games.

This is a very simple and minimal implementation that only maps RDRAM, if desired this could easily be extended to map the entirety of the N64 addressing space, although I don't know the exact specifics of what memory goes where.

I'm also not certain about the location of the code for this, I have it in main.c so that it get set after the pointer to RDRAM gets initialized, but is there a better location to do this?


m4xw:
whats the use case for this? never heard of those "network commands"

You want to expose it in and with https://github.com/libretro/mupen64plus-libretro-nx/blob/develop/mupen64plus-core/src/device/device.c (needs appropriate bit set tho for the address bus)

JoshuaEagles:
Here's the docs for the network commands, https://docs.libretro.com/development/retroarch/network-control-interface/, the two memory reading commands are the only ones that rely on these memory maps.

For my purposes, I'm trying to make it possible to use this core for Multiworld Ocarina of Time randomizers using Archipelago (https://archipelago.gg), the bsnes-mercury core has these memory maps setup and that allows it to used for SNES games in Archipelago. This would also enable other N64 games to work, and purposes other than Archipelago like inventory trackers.

I'll move it to device.c tonight. Thanks for taking a look at this.

JoshuaEagles:
I moved the memory map setup to device.c, placed it into init_device() which seems like a sane location, runs at almost the same point in execution as how I had it before and that method already deals with memory mappings. 

I did look into using the memory mapping struct for setting up the retroarch memory mappings, it seems doable, but I don't currently know how to test. The RDRAM memory map I've tested pretty carefully with memory offsets used by the Bizhawk script for connecting Ocarina of Time to Archipelago and the memory lines up, accounting for Bizhawk doing memory reads in little endian and applying a XOR 3 to the address before using it (haven't found the reason for that yet).

Not sure what you meant by needing an appropriate bit set for the address bus, looked around and I couldn't find anything that seemed related, sorry, not very familiar with this codebase.

m4xw:
See here
```

    struct mem_mapping mappings[] = {
        /* clear mappings */
        { 0x00000000, 0xffffffff, M64P_MEM_NOTHING, { NULL, RW(open_bus) } },
        /* memory map */
        { A(MM_RDRAM_DRAM, dram_size-1), M64P_MEM_RDRAM, { &dev->rdram, RW(rdram_dram) } },
        { A(MM_RDRAM_REGS, 0xfffff), M64P_MEM_RDRAMREG, { &dev->rdram, RW(rdram_regs) } },
        { A(MM_RSP_MEM, 0xffff), M64P_MEM_RSPMEM, { &dev->sp, RW(rsp_mem) } },
        { A(MM_RSP_REGS, 0xffff), M64P_MEM_RSPREG, { &dev->sp, RW(rsp_regs) } },
        { A(MM_RSP_REGS2, 0xffff), M64P_MEM_RSP, { &dev->sp, RW(rsp_regs2) } },
        { A(MM_DPC_REGS, 0xffff), M64P_MEM_DP, { &dev->dp, RW(dpc_regs) } },
        { A(MM_DPS_REGS, 0xffff), M64P_MEM_DPS, { &dev->dp, RW(dps_regs) } },
        { A(MM_MI_REGS, 0xffff), M64P_MEM_MI, { &dev->mi, RW(mi_regs) } },
        { A(MM_VI_REGS, 0xffff), M64P_MEM_VI, { &dev->vi, RW(vi_regs) } },
        { A(MM_AI_REGS, 0xffff), M64P_MEM_AI, { &dev->ai, RW(ai_regs) } },
        { A(MM_PI_REGS, 0xffff), M64P_MEM_PI, { &dev->pi, RW(pi_regs) } },
        { A(MM_RI_REGS, 0xffff), M64P_MEM_RI, { &dev->ri, RW(ri_regs) } },
        { A(MM_SI_REGS, 0xffff), M64P_MEM_SI, { &dev->si, RW(si_regs) } },
        { A(MM_DOM2_ADDR1, 0xffffff), M64P_MEM_NOTHING, { NULL, RW(open_bus) } },
        { A(MM_DD_ROM, 0x1ffffff), M64P_MEM_NOTHING, { NULL, RW(open_bus) } },
        { A(MM_DOM2_ADDR2, 0x1ffff), M64P_MEM_FLASHRAMSTAT, { &dev->cart, RW(cart_dom2)  } },
        { A(MM_IS_VIEWER, 0xfff), M64P_MEM_NOTHING, { &dev->is, RW(is_viewer) } },
        { A(MM_CART_ROM, rom_size-1), M64P_MEM_ROM, { &dev->cart.cart_rom, RW(cart_rom) } },
        { A(MM_PIF_MEM, 0xffff), M64P_MEM_PIF, { &dev->pif, RW(pif_mem) } }
    };
```
    
f.e. `#define MM_RDRAM_DRAM       UINT32_C(0x00000000)` is the base addr of dram

Cart DOM   (the cart image exposed at 0xB0000000) f.e. would be #define  UINT32_C(0x10000000) but that addr is in real 0xB0000000 so it needs the uncached memory bit set (0x10000000 | 0xA0000000 = 0xB0000000)

This works for all the mmio regions

m4xw:
(0xA0000000  is also uncached rdram)

JoshuaEagles:
Okay, to make sure I'm understanding this correctly (based on some of my own research as well):
- 0x80000000 - 0x9fffffff is the cached address space
- 0xA0000000 - 0xBfffffff is the uncached address space
- The rest of the address space is able to be decided by the game, so it probably doesn't make sense to map any of it for this

Both the cached and uncached address spaces represent the same data and the difference is just caching, so I think it makes sense for them to both be setup as retroarch memory maps so you can read/write either region interchangably.

I have a better idea of how this should look now, I'll get something implemented tomorrow that should have everything in the uncached and cached address spaces mapped to the respective regions.

JoshuaEagles:
Made a bunch of changes, I've set it up to construct a struct that defines the mappings to create for Retroarch, and then it uses that struct to create the mappings both starting at 0x80000000 and starting at 0xA0000000. Adding more, if ever needed, would only involve adding a new entry to that struct defining the mapping.

With this iteration I still kept it to only mapping RDRAM and the Cart ROM. Both of these I'm able to test and verify that the pointer is correct and reads and writes data correctly, or in the case of the Cart ROM it has the CONST flag set so writes aren't allowed. The other thing with most of the other data is there being some extra logic around reads and writes for most of it, which there isn't really a good way to represent through the Retroarch memory maps.

m4xw:
```
 * kuseg   0x00000000 - 0x7fffffff  User virtual mem,  mapped
 * kseg0   0x80000000 - 0x9fffffff  Physical memory, cached, unmapped
 * kseg1   0xa0000000 - 0xbfffffff  Physical memory, uncached, unmapped
 * kseg2   0xc0000000 - 0xffffffff  kernel-virtual,  mapped
 ```
 for you relevant are kseg0 and kseg1 only. You want to 1. create a mapping of the value of the mapping (kseg0 | addr) and another one for kseg1 | addr
Both are ultimately the same data. Not every addr is valid in kseg0 (cart dom iirc has to be uncached access, so the ROM only works | KSEG1) on hw. assuming i recall that correctly

**Comment by m4xw** [comment]
idaapi.add_segm(0, 0xA3F00000, 0xA3F00028, ".rdreg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4000000, 0xA4001000, ".sp.dmem", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4001000, 0xA4002000, ".sp.imem", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4040000, 0xA4080008, ".spreg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4100000, 0xA4100020, ".dpcreg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4200000, 0xA4200010, ".dpsreg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4300000, 0xA4300010, ".mireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4400000, 0xA4400038, ".vireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4500000, 0xA4500018, ".aireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4600000, 0xA4600034, ".pireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4700000, 0xA4700020, ".rireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA4800000, 0xA480001C, ".sireg", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA5000000, 0xA6000000, ".cartdom2addr1", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA6000000, 0xA8000000, ".cartdom1addr1", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xA8000000, 0xB0000000, ".cartdom2addr2", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xB0000000, 0xB0000000 + ROM_SIZE, ".cartdom1addr2", "DATA", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xBFC00000, 0xBFC007C0, ".pifrom", "CODE", ADDSEG_SPARSE)
idaapi.add_segm(0, 0xBFC007C0, 0xBFC00800, ".pifram", "CODE", ADDSEG_SPARSE)

Thats about the ones you want

**Comment by JoshuaEagles** [comment]
Looked around a decent amount and the only resource I could find that talked about which address ranges could be accessed with caching was this page: https://n64brew.dev/wiki/Memory_map, which states that only RDRAM supports it. So I have it setup so only RDRAM is available in KSEG0 and everything else is only available in KSEG1.

I have the entire address space mapped now, I tried to follow the mapping list you gave me, only two things are different from it. In your mapping list the sp mem is split into dmem and imem, but the code here doesn't seem to reflect that and they're both in dev->sp.mem based on the size of the mapping in init_device. The other place it differs is that the list you gave me doesn't have RSP_REGS2. I put it in since I had the mapping information for it, but I can easily remove it if it shouldn't be mapped here.

Also not sure which sections should be read only, a bunch of sections seem to have extra logic that runs when writing to them that wouldn't happen if written through Retroarch. For now I kept it to only ROMs being read only, since those seem like they should be for sure. I can easily mark others as read only as needed.

Did my best to test everything, although I can only verify for sure that the pointers for the N64 rom and DD rom are for sure since I can look at the roms in a hex editor, everything else I just made sure you could read from them correctly at both the start and end of the address ranges, and I double checked the code.

**Comment by m4xw** [comment]
I will check whats to do from there

**Comment by JoshuaEagles** [comment]
If there's any changes you want me to make or any other way I can help with this please let me know, although I assume you're just busy, so no worries.

**Comment by JoshuaEagles** [comment]
Hi, just checking in, haven't heard back in awhile, any updates on this?

**Comment by JoshuaEagles** [comment]
Ah, didn't know you were on vacation, sorry. Not a problem though, knowing you still intend to get to this when you have time is enough for me.

It's not quite only me using it, I have the client I'm using this for in a beta state and have some people testing it, but it's definitely still a very small number of users at this point.

**Comment by m4xw** [comment]
This is on the list for when i merge the gliden rebase

**Comment by m4xw** [comment]
Scheduled to go in next, i need a test case tho to try it

**Comment by JoshuaEagles** [comment]
Right, I can put something together to make it easy to test, at least for verifying stuff like RDRAM where its easy to read/write and see the correct output. I can get you that after work.

**Comment by JoshuaEagles** [comment]
Okay, so the way that you test this is by first enabling network commands in retroarch, which required Show Advanced Settings to be enabled (probably already have that) and then scrolling to the bottom of the network section of the settings and enabling it. 
image

Then, well I know how to do it on Linux, I'm not sure what OS you use or if you already have a preferred tool for sending data over udp, but you can use netcat to send commands to it. Steps for me are to first open a game on Retroarch (in this case I'm running Ocarina of Time 1.0) with network commands enabled, then running `nc -u 127.0.0.1 55355` to connect, and then you can use the READ_CORE_MEMORY and WRITE_CORE_MEMORY commands from there. Excerpt from the Retroarch docs:

```
READ_CORE_MEMORY <address> <number of bytes>
WRITE_CORE_MEMORY <address> <byte1> <byte2> ...
```

Then you can poke at values in memory and see stuff happen in the game. There's a sort of swizzle that the memory has when trying to read it so using a memory map like https://wiki.cloudmodding.com/oot/Save_Format isn't perfectly straightforward. Based on that documentation the current health is at `8011A600`, but you actually want to run that address through this function first to get the real address to use: `(address XOR 3) - (numberOfBytes - 1)`, which comes out to 8011A602. This is the address that Retroachievements lists as the address for this data, and bizhawk also reverses the swizzle in it's memory interface, so this seems to be expected behavior.

All that to say, if you load a file on a 1.0 OOT rom and Read/Write to 8011A602 you can get/set Link's health, setting it to 0 will kill him. I've tested this RDRAM behavior pretty extensively with my client I wrote (https://github.com/JoshuaEagles/OOT-RetroArch-Archipelago-Client) so I'm pretty confident about it. I can get you addresses to stuff in other games or something if you'd prefer, RetroAchievements has code notes available for a lot of games that are a good source for that.

I don't really know how to test most of the other parts, I was able to verify that the start of the ROM data matched what I expected when looking at the ROM in a hex editor, and I did something similar with the DD rom as well.

**Comment by m4xw** [comment]
Okay so all pretty basic.

I just noticed that this will run into issues with the recompiler cache if this is used to modify game code (rather  than .data section)

Is there a way we can have a callback or trigger? Might need to extend RA for that, i can take a look

**Comment by m4xw** [comment]
basically we need to cache flush after the frontend did the write

**Comment by JoshuaEagles** [comment]
Hey just to confirm, is that a change you want me to make? I'd need a bit more direction if so. If you're still looking into it not a problem, just checking.

**Comment by m4xw** [comment]
Sorry i was sick for a whole week and then i had another project... Feeling bad for the current delays. I dont need you to do anything right now. I just think this API was never designed for any kind of write operation and Ideally i talk to one of the people that wrote the integration in RA regarding this. I assume u want the write functionality in particular?

**Comment by m4xw** [comment]
If you dont need guarantee that the write memory works safely in all circumstances then its more or less okay as-is

**Comment by JoshuaEagles** [comment]
Hi, any updates on this? It'd be nice if we could merge this.

**Comment by m4xw** [comment]
Yes bunch of stuff will go in as soon i am done with the rebase, this included.

**Comment by JoshuaEagles** [comment]
Awesome, thanks for the update.

**Comment by m4xw** [comment]
I'd be fine with a squash and merge from here but one last thing i'd like if we can merge the mapping struct with the one inside of init_device

**Comment by JoshuaEagles** [comment]
Hm, I'm sure there's a way that could be done, but the way I've been looking at it the two structs serve pretty different purposes, even if they deal with a lot of the same data. 

Some of the pointers I used for the memory are slightly different from what's in init_device (eg for cart rom I used `dev->cart.cart_rom.rom` and the init device struct is instead `&dev->cart.cart_rom). So as far as I can tell merging the structs would require adding a separate pointer along with the isAccessibleCached and flags values that I need. 

The init_device struct also doesnt match in terms of rows, there's some rows that are in one but not the other. From a quick glance I can see the init device struct has a row referencing `MM_IS_VIEWER` which I don't have an equivalent for, and the retroarch struct has a row for `PIF_RAM` that doesn't have an equivalent. I don't personally need anything other than DRAM for what I'm using this for but I did put together the struct with the exact mappings you asked for above.

I guess I just want to double check that you do want that change, and if you do then how should I deal with the discrepancies between them? Should I just prefer the init_device struct?

**Comment by m4xw** [comment]
Man I wonder whats up with the notifications i miss yours all the times for some reason. Please just bug me if i dont respond within a day or two.

Fwiw probably easiest to make a macro. You are free to touch the device struct. Thats the "source of truth" for the device maps. I can maintain them going forward. Just try to not reformat it, splitting the args into a macro should work fine

**Comment by JoshuaEagles** [comment]
Hey sorry, been busy this past week, should have a chance to do this in the next couple days. For parts that differ between the mappings I have and the existing struct I'll just prefer the existing struct, I think that makes sense as the way to resolve the conflicts. So I'll just just add a few more points of data to it to provide the things I need. I can always make further changes after you look at it.

**Comment by m4xw** [comment]
Sure

**Comment by JoshuaEagles** [comment]
So update on this, rather strange issue I discovered while working on those refactors.

In short, my changes break RetroAchievements, and fixing that isn't trivial because for some strange reason they map the expansion pak memory to the start of KSEG0. This is where they define their memory regions for the N64, my understanding is the first two numbers in the structs are relative to the start of the pointer, and the 3rd is supposed to be relative to the console's address space. 
image
https://github.com/libretro/RetroArch/blob/65f76bd514ae26074ec5843799a4e85236eba440/deps/rcheevos/src/rcheevos/consoleinfo.c#L712

I asked them about this on their discord server, but they aren't very interested in fixing it. It would be a pretty large amount of work for them since the achievement sets rely on the memory addresses and they'd need to fix all the sets. They'd also need every emulator that implements support to update, as well as all users to update their emulators. So, even if they did want to fix it, it would take a super long time to do so.

That leaves this in a kinda weird spot, I think an okay solution here would be to map all of the memory starting at `0x0000 0000` instead of starting at KSEG0, and then to map the expansion pak memory a second time into the address that RetroAchievements expects. It's not a great solution, I don't love it, but I'd also not want to indefinitely delay this PR and that seems like the only way to work around this problem.

Thoughts?

Edit: talked to them a bit more and it does sound like in the long term they might be able to fix it, but they'd have to have a transition period where the old address is still mapped, so it'd likely be a very long time before the mappings could be done correctly. But, I think mapping it weird as a temporary thing sounds like a more reasonable compromise.

**Comment by m4xw** [comment]
Sounds like they use physical mappings.
```
#define	MIPS_PHYS_MASK			0x1fffffff

#define	MIPS_KSEG0_TO_PHYS(x)	((uintptr_t)(x) & MIPS_PHYS_MASK)
#define	MIPS_PHYS_TO_KSEG0(x)	((uintptr_t)(x) | (intptr_t)MIPS_KSEG0_START)
#define	MIPS_KSEG1_TO_PHYS(x)	((uintptr_t)(x) & MIPS_PHYS_MASK)
#define	MIPS_PHYS_TO_KSEG1(x)	((uintptr_t)(x) | (intptr_t)MIPS_KSEG1_START)

#define	MIPS_KSEG0_P(x)		(((intptr_t)(x) & ~MIPS_PHYS_MASK) == MIPS_KSEG0_START)
#define	MIPS_KSEG1_P(x)		(((intptr_t)(x) & ~MIPS_PHYS_MASK) == MIPS_KSEG1_START)
#define	MIPS_KSEG2_P(x)		((uintptr_t)MIPS_KSEG2_START <= (uintptr_t)(x))

/* memory map constants */
#define MM_RDRAM_DRAM       UINT32_C(0x00000000)
#define MM_RDRAM_REGS       UINT32_C(0x03f00000)

#define MM_RSP_MEM          UINT32_C(0x04000000)
#define MM_RSP_REGS         UINT32_C(0x04040000)
#define MM_RSP_REGS2        UINT32_C(0x04080000)
#define MM_DPC_REGS         UINT32_C(0x04100000)
#define MM_DPS_REGS         UINT32_C(0x04200000)
#define MM_MI_REGS          UINT32_C(0x04300000)
#define MM_VI_REGS          UINT32_C(0x04400000)
#define MM_AI_REGS          UINT32_C(0x04500000)
#define MM_PI_REGS          UINT32_C(0x04600000)
#define MM_RI_REGS          UINT32_C(0x04700000)
#define MM_SI_REGS          UINT32_C(0x04800000)

#define MM_DOM2_ADDR1       UINT32_C(0x05000000)
#define MM_DD_C2S_BUFFER    UINT32_C(0x05000000)
#define MM_DD_DS_BUFFER     UINT32_C(0x05000400)
#define MM_DD_REGS          UINT32_C(0x05000500)
#define MM_DD_MS_RAM        UINT32_C(0x05000580)
#define MM_DD_ROM           UINT32_C(0x06000000)

#define MM_DOM2_ADDR2       UINT32_C(0x08000000)

#define MM_CART_ROM         UINT32_C(0x10000000) /* dom1 addr2 */
#define MM_IS_VIEWER        UINT32_C(0x13ff0000) /* IS-Viewer */
#define MM_PIF_MEM          UINT32_C(0x1fc00000)
#define MM_CART_DOM3        UINT32_C(0x1fd00000) /* dom2 addr2 */
```

N64 games check for transfer pak by just accessing the memory and see if it blows iirc, but that might vary

**Comment by m4xw** [comment]
I think as a workaround we can just add the map with MIPS_PHYS_MASK additionally? But its quite janky

**Comment by JoshuaEagles** [comment]
I don't think it would even need the PHYS mask, I have to manually add KSEG0/1 to the addresses I'm mapping from the struct. The annoying part is that I can't map stuff to KSEG0/1 at all anymore since the expansion pak memory needs to go at the very start of KSEG0. I think code-wise it should be fine.

I think I'll just go ahead and implement it as physical mappings and leave a comment to make it clear why the expansion pak is mapped weirdly. I might also make an issue on their github so I can link to it from the code

**Comment by m4xw** [comment]
In practice there is no meaningful overlap iirc, you can just expose both, any regs get accessed with uncached memory, rdram 0x80...  and 0xA range as well, their mapping is nonsense

**Comment by m4xw** [comment]
for your purpose 0x80000.. range is the 0xA000..  and 0x000... range, there is no difference
For expansion pak i guess they just mask out the range but thats all very snowflake stuff

File: mupen64plus-core/src/device/device.c [hunk 1]
```diff
@@ -77,6 +83,82 @@ static void get_pi_dma_handler(struct cart* cart, struct dd_controller* dd, uint
...
+    size_t start;
+    size_t len;
+    void* ptr;
+    bool isAccessibleCached;
+    uint64_t flags;
+};
+
+void setup_retroarch_memory_map(struct device* dev, size_t rom_size, size_t dd_rom_size) {
+    struct n64_to_retroarch_memory_map n64_to_retroarch_mappings[] = {
+        { MM_RDRAM_DRAM,             0x3efffff,    dev->rdram.dram,        true,  RETRO_MEMDESC_SYSTEM_RAM },
```
m4xw: 0x3efffff should be -> rdram_size

mupen64plus-core/src/device/device.c
```diff
@@ -77,6 +83,82 @@ static void get_pi_dma_handler(struct cart* cart, struct dd_controller* dd, uint
...
+    for (int i = 0; i < n64_to_retroarch_mapping_count; i++) {
+        struct n64_to_retroarch_memory_map mapping = n64_to_retroarch_mappings[i];
+
+        if (mapping.ptr == NULL) {
+            continue;
+        }
+
+        if (mapping.isAccessibleCached) {
+            descs[i].ptr = mapping.ptr;
+            descs[i].start = R4300_KSEG0 + mapping.start;
```
m4xw: should be OR |

mupen64plus-core/src/device/device.c
```diff
@@ -77,6 +83,82 @@ static void get_pi_dma_handler(struct cart* cart, struct dd_controller* dd, uint
...
+            descs[i].len = mapping.len;
+            descs[i].flags = mapping.flags;
+        }
+
+        descs[i + n64_to_retroarch_mapping_count].ptr = mapping.ptr;
+        descs[i + n64_to_retroarch_mapping_count].start = R4300_KSEG1 + mapping.start;
```
m4xw: should be OR |

mupen64plus-core/src/device/device.c
m4xw: mapping size they should use to determinate if expansion pak is used

m4xw (changes requested):
formal review

m4xw:
grafik

m4xw:
Just noticed u wont be able to include it in the table if u dont reference it by ptr, so maybe add another column with &rdram_size, also u need to make sure this runs after load game, in fact i wonder if we might need to defer the init to the first retro_run call (cheevos already has to deal with that for mupen)

mupen64plus-core/src/device/device.c
JoshuaEagles: Since I need to split the mapping into 3 to match how RetroAchievements maps it, I can just not map the expansion pack when the size is smaller, should be pretty straightforward

JoshuaEagles:
I have quite a lot of changes locally that I haven't pushed yet, I have the mappings using the same struct as init_device, although doing that led to a significant jump in complexity in the setup_retroarch_memory_map method. Now I'm working on setting up those RetroAchievements mappings, and then I'll try and clean it up as much as I can. Not had a lot of time for this recently, but I'm pretty close at this point

m4xw:
> I have quite a lot of changes locally that I haven't pushed yet, I have the mappings using the same struct as init_device, although doing that led to a significant jump in complexity in the setup_retroarch_memory_map method. Now I'm working on setting up those RetroAchievements mappings, and then I'll try and clean it up as much as I can. Not had a lot of time for this recently, but I'm pretty close at this point

This is inaccurate in terms of hardware and we do not cater for them, they do for us so i'd prefer you dont bother with that and they should just fix it. I can talk to them directly

m4xw:
Primary use case i see for this PR is modding frameworks and other external interaction, not cheevos

mupen64plus-core/src/device/device.c
m4xw: Just map them all like i said in the other comment.
*they* need to use the mapping size to detect the pak, not us.

JoshuaEagles:
I agree with you, don't get me wrong, but from the convo I had with them it would take an extremely long time to change it since they'd need a transition period to allow people to update their emulators. You can try talking to them but I don't think it would have a different outcome.

The way I see it, breaking them isn't an option, but I'd also like this merged sometime, ideally. Ultimately you're the maintainer, if you don't see mapping it incorrectly as an option (which is fair) then I think that just means this PR would have to just wait until they fix their code, however long that takes

m4xw:
its just affecting old cores on new retroarch, i consider this a minor thing. dont get me wrong too, i probably care less about it than i sound like. my primary issue is i'd want 0x8 range all exposed because thats what anything that uses the integration would actually use (even tho its equivalent to the uncached range). I mean technically this whole thing is cursed since any memory modification might not trigger any interrupts or cache invalidation in case of it patching parts of the code.

Their fix is a one liner

m4xw:
But given this was never supported before (no maps) and cheevos already worked via the poking api, why not just have them use the current way for now? As in consider n64 as no map support and use current codepath

JoshuaEagles:
The way rcheevos works it checks for maps and gets the memory a different way only if they don't exist. I had to go pretty deep into their code to even understand what the issue was, so I saw that it worked like that. 

I suppose they could add special case logic for specifically N64, but even if they did that then a new Retroarch stable version would have to come out with an updated rcheevos, and users would need time to update before this would be safe to merge, so pretty much the same issue

m4xw:
> The way rcheevos works it checks for maps and gets the memory a different way only if they don't exist. I had to go pretty deep into their code to even understand what the issue was, so I saw that it worked like that.
> 
> I suppose they could add special case logic for specifically N64, but even if they did that then a new Retroarch stable version would have to come out with an updated rcheevos, and users would need time to update before this would be safe to merge, so pretty much the same issue

triviality to roll out for them and that wouldnt be a first for mupen either (they already defer init for me after the first frame). the other cores are technically EOL anyway so its only about catering to old cores

m4xw:
Now that i think about it, might just make it a core option and call it a day?

m4xw:
Need to check if the cb is available early enough to nope out of the memory setup however and change it to allow cheevo compatible and accurate via the setting

JoshuaEagles:
Yeah a core option would work here, defaulting to cheevo compatible would probably make sense. If that route works here could you point me to a good spot to use as a reference for how to implement that?

m4xw:
you need to add it to mupen64plus common header (for the extern of the variable), libretro core options header for the strings, then just query it like the other ones in update_variables in libretro.c i think it was and the actual variable for the core option (that u externed earlier in the header) should go somewhere at the top of libretro.c next to all the others (put it at the end), you need to check that it happens prior tho to initialising the map, i would need to doublecheck the api spec.
Make sure it initialises to 0 and thats the default setting you want

m4xw:
Here is a example commit for a gliden config example
https://github.com/libretro/mupen64plus-libretro-nx/commit/3a676196500545b637b83cb19fb393d2359e1f9d

m4xw:
Lets call the option Memory Map Layout: Cheevos Legacy=0, Accurate=1

JoshuaEagles:
Alright, this looks perfect for figuring this out. Thanks!

What're you picturing exactly for how the Cheevos Legacy option should work? Should I just not do memory maps at all, or map it to the 0x0 range? The code I have locally has gotten pretty complex due to using that other struct, so I'm a little wary of adding to that further by mapping everything to a different location depending on that setting

m4xw:
i assume we map all ranges but only the 0x8 range if expansion pak?

m4xw:
I talked to them and the concern is solved for the time being, please go ahead and do the review requested changes.
Some notes will follow

JoshuaEagles:
Oh wow, interesting, I'm curious what changed since I talked to them. I should have some time tonight so I'll try getting what I have locally ready to commit and push

m4xw:
Just make sure we map 0xA, 0x8 and 0x0 range (phys) basically
grafik

JoshuaEagles:
Okay, I have something working at this point. Setting select and disconnect like that does appear to work, RetroAchievements is detecting all of the memory correctly even for expansion pak games like Majora's Mask. I'll be honest though, I don't fully understand how those values were derived. I think it would theoretically be possible to map to 0x0, 0xA, and 0x8 using one mapping via select and disconnect, but these values don't allow for that (they only map 0x0 and 0x8) and if I don't understand the reasoning behind those values it's difficult for me to look for new ones. I have the KSEG1 mapping of RDRAM as a separate mapping for now, if it can be consolidated then that would be nice though.

While testing I also noticed that a bunch of the pointers that are needed only get set after the mappings struct is created, for example, RDRAM, cart rom, and dd rom, anything that calls mem_base_u32 appears to be like this. The way I have this structured has a decent amount of duplication in it, but I think trying to reduce that duplication would just make this code harder to understand. I'm honestly thinking having the second struct was preferable, the code was much leaner and simpler overall, the fact that the other struct doesn't actually really work for what I need makes the code to use it way more complicated than it should be, but I'll leave it up to you.

I've pushed an updated version of the core, I've re-tested everything although I have very little means to verify the data for anything except RDRAM, cart rom, and DD rom, but those 3 I can at least say expose the right memory. This new version does still work for what I need it for, but that only involves RDRAM.

Let me know what you think. I'll implement any feedback you have. If you do prefer how it was before I also don't mind reverting this to that version.

JoshuaEagles:
Hi @m4xw , any updates on this?

m4xw:
will check on monday

m4xw:
Everything pushed on your end? On a quick look seems fine, might be okay to merge

JoshuaEagles:
Yeah, my push a month ago was my most recent work on this. I tested it as thoroughly as I could, so if it looks good to you then you should be good to merge

m4xw:
We will see what blows the hard way

PR #547, opened 2024-08-13
iOS/tvOS: move build flags inside makefile

warmenhoven:


warmenhoven:
When I created #538 I was just building manually and didn't even see that the gitlab CI was overriding it so that HAVE_NEON was never actually getting set.

PR #559, opened 2024-11-13
Upload audio once per frame

sonninnos:
Copypasted the same solid code used in a few cores, so that audio is uploaded once per frame after video callback, and not twice in between.


golivax:
@sonninnos Thanks for this PR. I really wanted to test it out. Is there any chance you can share a Windows build? Thanks! [I am not a dev in this project]

sonninnos:
[mupen64plus_next_libretro.zip](https://github.com/user-attachments/files/19152709/mupen64plus_next_libretro.zip)

sonninnos:
So how was the testing?

Rebased and tried to resolve the merge conflict but now there are new ones..

PR #568, opened 2025-02-28
Add possibility to snap analog stick controller angle

halest:
Originally a fix for a different core I decided to also write support for Mupen. 
The original discussion can be found [here](https://github.com/libretro/parallel-n64/issues/792)

As stated there:
What

Many emulators have a setting to snap the control stick to cardinal directions to simulate a control stick like on the N64 or GameCube controller.
Usually this includes a setting, which offsets to snap to the nearest cardinal direction.
For example the setting is set to "5", angles from 85 to 95 would snap to 90.

Why

Closer emulation of classic controller behaviour
Speedrunning and such
Better support for circular controllers


halest:
Overnight I had an idea on how to improve this design. I will close the request and open a new one once I am finished implementing it

PR #570, opened 2025-03-09
Added support for multiple transferpak games

JPZV:
## Description

This PR adds support for loading different GB games on transferpak for each controller. This allows, for example, to play Pokemon Red for P1, Blue for P2, and Yellow for P3 on Pokemon Stadium

## How it works (For the End-User)

Until now, you had to have both the .gb and .sav files alongside the N64 ROM (and with the same name), having this structure:

```
roms/
  n64/
    pokemon stadium.z64
    pokemon stadium.z64.gb
    pokemon stadium.z64.sav
```

But now you have to create a .pak folder and inside place the .gb and .sav files for every controller, naming them from 1.gb to 4.gb. For example:

```
roms/
  n64/
    pokemon stadium.z64
    pokemon stadium.z64.pak/
      1.gb
      1.sav
      2.gb
      2.sav
      3.gb
      3.sav
      4.gb
      4.sav
```

Also, you can combine both the old and the new methods (the newer one has higher priority). For example:

```
roms/
  n64/
    pokemon stadium.z64
    pokemon stadium.z64.gb
    pokemon stadium.z64.sav
    pokemon stadium.z64.pak/
      1.gb
      1.sav
```

With that, only the first player will have a different ROM while other players will use the same _pokemon stadium.z64.gb_ ROM.

## How it's implemented (for Libretro maintainers)

Now the global variables `retro_transferpak_rom_path` and `retro_transferpak_ram_path` are string arrays of fixed length of 4 (one for each controller). Then, on startup, the core will search for each .gb and .sav file (naming from 1 to 4) inside the `{GAME_FILE_NAME}.pak` folder. If there is no .gb or .sav file for that controller, then it'll rollback to the old method and, in case there are those files, it'll assign those files for that specific controller.

So, in pseudo code, it'd be something like this:

```c
rom_path[4], ram_path[4] as array of char*;
for (i = 0 to 4)
{
  if (file_exists("{GAME_FILENAME}.pak/{i}.gb")
  {
    rom_path[i] = "{GAME_FILENAME}.pak/{i}.gb";
    if (file_exists("{GAME_FILENAME}.pak/{i}.sav")
    {
      ram_path[i] = "{GAME_FILENAME}.pak/{i}.sav";
    }
  }
  if (empty(rom_path[i]) or empty(ram_path[i]))
  {
    if (file_exists("{GAME_FILENAME}.gb")
    {
      rom_path[i] = "{GAME_FILENAME}.gb";
      if (file_exists("{GAME_FILENAME}.sav")
      {
        ram_path[i] = "{GAME_FILENAME}.sav";
      }
    }
  }
}
```

### DIFF
--- a/custom/mupen64plus-next_common.h
+++ b/custom/mupen64plus-next_common.h
@@ -76,8 +76,8 @@ extern char* retro_dd_path_img;
 extern char* retro_dd_path_rom;
 
 // Other Subsystems
-extern char* retro_transferpak_rom_path;
-extern char* retro_transferpak_ram_path;
+extern char* retro_transferpak_rom_path[4];
+extern char* retro_transferpak_ram_path[4];
 
 // Threaded GL Callback
 extern void gln64_thr_gl_invoke_command_loop();
--- a/libretro/libretro.c
+++ b/libretro/libretro.c
@@ -159,8 +159,8 @@ char* retro_dd_path_img = NULL;
 char* retro_dd_path_rom = NULL;
 
 // Other Subsystems
-char* retro_transferpak_rom_path = NULL;
-char* retro_transferpak_ram_path = NULL;
+char* retro_transferpak_rom_path[4] = { NULL };
+char* retro_transferpak_ram_path[4] = { NULL };
 
 uint32_t CoreOptionCategoriesSupported = 0;
 uint32_t CoreOptionUpdateDisplayCbSupported = 0;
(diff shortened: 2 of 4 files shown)


m4xw:
Mustve missed this PR, will check

m4xw:
probably part of the next batch of merges

Eulenberg:
Thank you for putting this pull request together and for your work on this feature. It would be a huge feature for my upcoming event, a local pokemon tournament where players bring their own save files.

Since this was planned for merging around a year ago, I wanted to gently check in on its current status. Are there any remaining blockers, feedback, or open items that need to be addressed before this can be merged? Please let us know if any help is needed to get this across the finish line.

PR #571, opened 2025-03-14
Threaded emscripten support

JoeOsborn:
This patch (on top of emscripten_testing, not develop) allows for pthread compilation of mupen64.  Unfortunately as long as libco is still in use it must be built with ASYNC=1 as well as pthread=(POOL_SIZE).


m4xw:
I need to resync my emscripten branch its a bit out of date again..

Makefile
```diff
@@ -526,7 +526,10 @@ else ifeq ($(platform), emscripten)
    CC = emcc
    CXX = em++
    HAVE_NEON = 0
-
+   ifneq ($(pthread),0)
+     CPUFLAGS += -pthread
+     LDFLAGS += -lpthread
```
orbea: Drive by review:

The -pthread flag is both a linker flag and a cflag, you shouldn't use -lpthread directly and instead rely on the compiler to set it with -pthread.

JoeOsborn:
This is rebased on upstream/develop and adds a fix for savestates under emscripten-with-pthreads.

PR #576, opened 2025-05-29
Hide "EnableCopyColorFromRDRAM" for non-GLideN64 plugins

bslenul:
Forgot to actually open the PR... :x I have it ready for 4 months now, sorry!

Without a description with `(GLN64)` in it the option isn't hidden when using other renderers.


m4xw:
Forgot about that, its some legacy behaviour from way before new core options...

PR #581, opened 2025-08-01
port kev4cards' excellent N64 analog algorithm from ares

hunterk:
This produces near-perfect analog stick behavior on modern controllers:
image


hizzlekizzle:
@m4xw if it makes it any more attractive to merge, I moved the new behavior to a (default off) core option and cleaned up the formatting a bit.

PR #588, opened 2025-08-22
Compile fix for libnx 4.9.0. Update mmap with the new virtmem API.

rwesterlund:
I recently set up a development environment to build RetroArch for Switch. I could not compile this core so I looked into it and made a compile fix for the latest version if you consider updating to that. I don't know which libnx version is supposed to be used to compile the current develop.

This affects the the DynaRec compiler that I am not too familiar with but because it wants to allocate executable memory I picked an appropriate alloc routine for that. I have tested it with OoT and Banjo-Kazooie (previously uncompiled) and they ran without crashes but no longer sessions yet.

### DIFF
--- a/switch/mman.h
+++ b/switch/mman.h
@@ -29,7 +29,9 @@ static inline void *mmap(void *addr, size_t len, int prot, int flags, int fd, of
     (void)offset;
 
     size_t size = (len + 0xFFF) &~ 0xFFF;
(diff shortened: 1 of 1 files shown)


m4xw:
We are shipping https://git.m4xw.net/Switch/RetroArch/libnx/-/commits/deprecated-features

PR #609, opened 2025-12-15
Add webOS to Makefile / CI

cscd98:
Fixes building on webOS which needs GLES (not GL).

It also changes the TARGET name to add gles 2 and gles 3 to the filename, because the gitlab buildbot was overwriting the previous compiled version, meaning gles 2 was never available to download.

A PR to update the info files has already been merged.

Now they can both be listed for download

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -46,6 +46,9 @@ include:
   # tvOS
   - project: 'libretro-infrastructure/ci-templates'
     file: '/tvos-arm64.yml'
+  # webOS 32-bit (LGTV)
+  - project: 'libretro-infrastructure/ci-templates'
+    file: '/webos-make.yml'
 
 # Stages
 stages:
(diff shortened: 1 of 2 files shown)


cscd98:
@m4xw please could you look at merging this PR? :+1:

LibretroAdmin:
Hi @m4xw , asking if this can be merged

m4xw:
> @m4xw please could you look at merging this PR? 👍

if its a dedicated platform with dedicated targets on ci/cd, at least add optimized cpu flags

cscd98:
@m4xw it inherits flags:

```
-Wno-discarded-qualifiers -O3 -DNDEBUG -fsigned-char -ffast-math -fno-strict-aliasing -fomit-frame-pointer -fvisibility=hidden -fcommon
```

Oh and I forgot the toolchain also sets its own flags, which is done for compatibility with webOS 1, so -mcpu is always cortex-a9

```
echo | arm-webos-linux-gnueabi-gcc -Q --help=target | grep -E 'mcpu|mtune|mfpu'
  -mcpu=                      		cortex-a9
  -mfpu=                      		neon-fp16
  -mtune=
```

PR #613, opened 2026-01-10
apple: use audiotoolbox for resampling

warmenhoven:
Slightly more performant, slightly better audio quality on apple. Avoids int/float conversion as well.


warmenhoven:
https://git.libretro.com/libretro/mupen64plus-libretro-nx/-/pipelines/39552

PR #617, opened 2026-02-07
Implement mouse mode

pstef:


m4xw:
Good set of PR's will merge them in one go as soon this is ready

pstef:
I consider this ready so if you see any deficiencies, could you tell me what they are?
I plan to send a PR that will add another feature on top of this one, but it may deserve a separate discussion.

m4xw:
primary concern is it conflicts with the input change pr's that are currently open (but they need to be re-done anyway), also conflicts with local changes, so not sure what order to merge them in, the other pr's i am kinda waiting forever on

PR #628, opened 2026-04-26
add linux-aarch64 build

theofficialgman:
this is untested on the libretro CI. Documentation is sparse on how to actually test these changes. If anyone could provide details on that process please do.


theofficialgman:
tested working. ready for merge.

WizzardSK:
I see that I made duplicate https://github.com/libretro/mupen64plus-libretro-nx/pull/639 but I also tested it and it works, so it is safe to merge.

PR #632, opened 2026-05-06
Apple Silicon (macOS arm64) dynarec port

### LINKED ISSUE #561: no dynarec on macOS 11?
is there no Dynamic recompilier available on macOS 11? or any version of macOS? The [Mupen64Plus-Next libretro docs](https://docs.libretro.com/library/mupen64plus/#core-options) page says that "Dynamic recompilier is not available on all platforms" but doesn't mention which platforms. Can the docs be updated to mention which platforms are supported?

RetroArch Metal 1.19.1
Mupen64Plus-Next 2.7-Vulkan 4249e39

AurisDSP:
## Summary

Brings the aarch64 dynarec online on macOS Apple Silicon (M1/M2/M3/M4). Resolves #561.

The host code ports cleanly with a handful of platform shims:

* **Build (Makefile):** enable `WITH_DYNAREC=aarch64` on `osx` for native arm64; keep `-DNO_ASM` only when no dynarec backend is in use.

* **Assembly (linkage_arm64.S):** Apple's clang assembler treats `;` as a comment delimiter, so the GAS-style cpp macros that emit multiple directives on one line don't work. Switched `GLOBAL_FUNCTION` / `LOCAL_FUNCTION` / `GLOBAL_VARIABLE` to assembler-side `.macro` definitions on Apple, with an `_`-prefixed Mach-O symbol and `.private_extern`. Added an `ADRP_LO12` macro that selects between `:lo12:` (GAS) and `@PAGE`/`@PAGEOFF` (Apple) syntax. Internal `b.ne new_dyna_stop` was rerouted through a local `L`-label so Apple's assembler doesn't reject the conditional branch's external target.

* **JIT memory (new_dynarec.c):** macOS Hardened Runtime forbids RWX in a single mapping, so on Apple we mmap the cache with `MAP_JIT` and toggle protection per-thread via `pthread_jit_write_protect_np()`. To handle `set_jump_target()` being called from inside `new_recompile_block` (where the page is already in write mode), the BEGIN/END pair nests via a per-thread depth counter (\`_Thread_local\`); END clamps to zero so `cache_flush()` called from invalidation paths still lands the page in execute mode.

* **W^X helpers (new_dynarec.h):** \`M64P_JIT_WRITE_BEGIN\` / \`M64P_JIT_WRITE_END\` macros wrap \`pthread_jit_write_protect_np\` with the nesting counter on Apple, no-ops elsewhere. The counter is \`_Thread_local\` since the underlying syscall is per-thread.

* **Cache flush (assem_arm64.c):** on Apple, replaced the manual \`dc civac\` / \`ic ivau\` loop with \`sys_icache_invalidate()\` and a paired \`M64P_JIT_WRITE_END\`. \`arch_init()\` and \`set_jump_target()\` now wrap their writes with BEGIN/END.

* **Portability (custom/dependencies):**
  * \`pngpriv.h\`'s \`TARGET_OS_MAC\` branch pulls in \`<fp.h>\` (a Mac OS 9 header), so it's been excluded on modern macOS.
  * \`zutil.h\`'s \`MACOS\` branch redefines \`fdopen\`, which collides with Apple's stdio.h \`__memcpy_chk\`-style macros; skip the redefinition on \`__APPLE__\`.

The second commit is hardening that fell out of an audit while looking at the Apple AudioConverter path — same fixes apply to the non-Apple sinc-resampler branch (NULL checks after malloc, zero-frequency guard, oversized-input truncation, drain-loop break on \`audio_batch_cb\` returning 0).

## What works

- Mario Kart 64 boots and plays with audio + video on M-series hardware (\`cached_interpreter\` and \`dynamic_recompiler\` both work; \`angrylion\` + HLE RSP avoids a paraLLEl-RDP / MoltenVK threading bug we'll file separately).
- Apple Silicon dylib loads cleanly with both video drivers (\`glcore\` and \`vulkan\` via MoltenVK).
- All Apple-specific code is guarded behind \`__APPLE__\` so Linux / Windows / Android paths are unchanged.

## What doesn't work yet (out of scope)

Some homebrew ROMs (e.g. *BikeRace '98*, *Wet Dreams* POM '99) hit a SIGSEGV inside generated code with \`LR=1\`-shaped corruption. The crash isn't macOS-specific — it reproduces on this branch's tree on Linux with the same recompiler — so it's a pre-existing aarch64 codegen edge case rather than something the macOS port introduced. Worth filing separately.

## Tested on

- M4 MacBook Air, macOS 26.x, RetroArch 1.22.2

## Test plan

- [x] Cached interpreter mode runs Mario Kart 64 end-to-end
- [x] Dynamic recompiler mode runs Mario Kart 64 end-to-end (boot, audio, video, input)
- [x] Build with \`make platform=osx\` succeeds clean
- [x] Linux paths unchanged (all changes are \`__APPLE__\`-guarded)


m4xw:
This will take some time to review

m4xw:
So, i'd rather have it use the w^x codepaths of mupen, that will have higher performance too, you can check how libnx does it  (buffer in .text is also to lessen distance for more perf), but the dual emitter mode with base rx != base rw is what you need.
Remapping is too costly (arguable if that will matter on mac itself), so i'd only use it if theres no way to have a RW mirror page of a RX buffer on Mac

Issue #619, opened 2026-02-21
Renderer options.


obsgamecapture:
It would be neat to have renderer option for Vulkan, OpenGL, Software etc.

So what i like to do is keep my Retroarch renderer with OpenGL to not it be captured by game capture and have the emulator be in Vulkan for it to be captured. This works with PS1 emulator for example, but not with N64.

Would be a nice option to have.

checktext00:
You can force Vulkan rendering for this core by choosing ParaLLEl-RDP as the RDP Plugin. Otherwise **GL**ideN64 forces Open**GL**

obsgamecapture:
Thanks.

PR #633, opened 2026-05-08
Specify ParaLLEl-RDP = Vulkan, GLideN64 = OpenGL

checktext00:
Makes clear that ParaLLEl-RDP forces Vulkan and that GLideN64 forces OpenGL


m4xw:
I dont think thats really needed, it will just pick whatever suitable driver

checktext00:
@m4xw it was meant to solve user confusion in issues like these: https://github.com/libretro/mupen64plus-libretro-nx/issues/619

PR #636, opened 2026-05-25
webOS: fix build on 64-bit and use of TEXTURE_EXTERNAL

cscd98:
Currently adding aarch64 support and needs a little fix.

I also noticed it will set eglImage and eglImageFramebuffer to true, this causes a segfault later in the mali driver as it cannot bind a external texture without backing.


cscd98:
@m4xw could you have a look / merge this PR please :+1:

PR #641, opened 2026-08-03
libretro: add vfs aware file checks in retro_load_game

cscd98:
This adds vfs aware path checks to libretro.c.

The rom itself doesn't need it as the core has info->need_fullpath = false


m4xw:
Insufficient for 64dd system files (basically bios etc)

PR #644, opened 2026-08-30
libretro: support booting a bare 64DD disk image

XenuIsWatching:
The core has long been able to boot a 64DD disk with no cartridge, but none of it was reachable: emu_step_load_data() always issued M64CMD_ROM_OPEN, so a disk was rejected by is_valid_rom() and M64CMD_DISK_OPEN was never sent.

 - Detect a disk in retro_load_game(), by .ndd extension, the two canonical dump sizes, or the JP/US region word in either endianness. Carts are ruled out first, so any cartridge takes the path it took before.

 - Supply g_media_loader.get_dd_disk. Nothing populated it, so open_disk() could never find a disk.

 - Send M64CMD_DISK_CLOSE, not M64CMD_ROM_CLOSE, when a disk is open. The core rejects ROM_CLOSE with no ROM open, failing the next load.

 - Skip the content copy on the disk path, since DISK_OPEN takes no data.

 - Add ndd to valid_extensions. need_fullpath stays false; changing it would break in-memory and archived cartridge loading.

Also fixes an out-of-bounds read: the num_info == 1 branch of RETRO_GAME_TYPE_DD fell through to info[1] on a one-element array.

Two core changes come with it. Only the first is required:

device.c: with no cartridge rom_size is 0 and rom_size-1 underflows to SIZE_MAX. The resulting mapping end is harmless and matches upstream, but setup_retroarch_memory_map() exports the entry unconditionally, handing the frontend a descriptor with a NULL pointer and a SIZE_MAX length.

dd/disk.c: an unrelated bug, not needed to boot a disk. scan_and_expand_disk_format() stored the loop index where both uses want the block number, so system data found in block 8-11 yields an offset pointing at block 4-7. It fails silently, returning the wrong region. The development test had the same confusion: 2 and 3 match by coincidence, while 10 and 11 never match an index running 0-7. Every intact dump keeps four copies, at blocks 0,1,8,9 or 2,3,10,11, and the scan takes the first, so this is only reachable once blocks 0-3 are unreadable. 15 disks checked, none affected.

### DIFF
--- a/.github/workflows/retroxr-release.yml
+++ b/.github/workflows/retroxr-release.yml
@@ -69,9 +69,15 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y g++-mingw-w64-x86-64 nasm binutils make zip
 
+      # No `submodules: recursive`, unlike the other retroXR core forks. This
+      # repository has no .gitmodules at all -- everything it vendors came in
+      # through git subrepo, which commits the trees -- and it carries one
+      # leftover gitlink, mupen64plus-rsp-paraLLEl/lightning/gnulib, with no
+      # entry to match. Asking for submodules therefore cannot succeed on any
+      # ref: `fatal: No url found for submodule path ... in .gitmodules`,
+      # exit 128, before a line is compiled. Nothing is missing without it.
       - uses: actions/checkout@v6
         with:
-          submodules: recursive
           ref: ${{ github.event.inputs.ref || github.ref }}
 
       # CC/CXX are passed on the command line rather than left to the makefile's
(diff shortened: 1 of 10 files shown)


XenuIsWatching:
Verified on Windows and Android  Mario Artist and Kyojin no Doshin boot from a bare .ndd, F-Zero X keeps its Expansion Kit through the sidecar, and plain cartridges are unaffected. I had a blank screen rather than a failed load with GLideN64 which renders nothing on a cartridge-less boot, while ParaLLEl-RDP and angrylion both render it correctly... i'm not sure what that was and i didn't chase it down

XenuIsWatching:
some of these files are are also in the upstream as well... but i dunno which the process is here for those, unless if you want those in a seperate PR for that one

XenuIsWatching:
it seems that some other commits keep getting stacked on here. This also came from a wrong branch pointer (althrough the first commit is correct), since GH won't let me repoint the branch, i just have to close and reopen :/

